PR #8, opened 2020-02-05
List previously created wikis

edg2s:


MatmaRex:
I'm not sure how I feel about the stylesheet… Styling communicates a certain expectation of reliability and stability, and no styling communicates exactly what I wanted to communicate ;)

edg2s:
It also communicated to other developers that they should invest their time in making it better :)

PR #16, opened 2020-02-06
Automatically detect server path

### LINKED ISSUE #5: Allow installation of patchdemo in a sub folder
Currently it is assumed that you are installing in `/var/www` on an otherwise empty VM. You may want to install in `/var/www/patchdemo`, e.g. for local developement.

edg2s:
Also make all links relative to the current path.

Fixes #5


MatmaRex:
Ehhh okay whatever.

PR #29, opened 2020-03-11
OAuth support

### LINKED ISSUE #25: Use Wikimedia OAuth for our public instance
We could then tie each instance to a username, which would help us keep an eye on abuse of the system. We could also allow users to delete their own instances.

edg2s:
When enabled, only authenticated users can create new wikis, and the creator of each wiki is stored and displayed in the list.

Fixes #25

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-config.json
+config.php
 wikis
 composer
 composer.lock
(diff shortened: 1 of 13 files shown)


edg2s:
-1 I need to make config.json unserveable!

PR #38, opened 2020-03-31
Link the logo to the main page

MatmaRex:
Particularly useful when you want to come back from new.php.

### DIFF
--- a/includes.php
+++ b/includes.php
@@ -20,7 +20,7 @@ echo '<!DOCTYPE html>
 		<link rel="stylesheet" href="index.css">
 	</head>
 	<body>
-		<a href="."><h1>Patch demo</h1></a>';
+		<h1><a href=".">Patch demo</a></h1>';
 
 include_once 'oauth.php';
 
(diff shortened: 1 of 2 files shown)


includes.php
```diff
@@ -20,7 +20,7 @@
 		<link rel="stylesheet" href="index.css">
 	</head>
 	<body>
-		<h1>Patch demo</h1>';
+		<a href="."><h1>Patch demo</h1></a>';
```
edg2s, line 23: Shouldn't put block inside inline. Put the link inside the heading, then change the CSS to target `h1 > a` instead of `h1`. Remove the link colour and underline while there.

edg2s:
+1, made changes

includes.php
MatmaRex: For the record, this way is not incorrect though. `<a>` is not inline but "transparent", and it may contain anything, including block elements.

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element
https://html.spec.whatwg.org/multipage/dom.html#transparent

PR #43, opened 2020-04-07
Add support for FlaggedRevs

DannyS712:
So that patches can be tested


MatmaRex:
I am not sure how I feel about this, this extension is not included by default with MediaWiki and it's not enabled on most Wikimedia wikis. But I added the DiscussionTools extension, so I feel like I don't really have an argument here. Maybe there should be a way to enable/disable extensions per wiki.

PR #44, opened 2020-04-08
Link to user pages

edg2s:


MatmaRex:
Everything here should be HTML-escaped, as usernames allow `&` (not `<` `>` though luckily), but I guess it's alreayd wrong, so we can fix that later.

Issue #19, opened 2020-02-23
Provide a way to override configs


edg2s:
This would allow demonstrations of features hidden behind feature flags, for example.

We don't want to let people add arbitrary PHP to local settings, so we'd probably need to do something JSON-based instead.

DemianX0:
@edg2s Do you need to change the flags after the demo was deployed or the flags could be part of the patch?

edg2s:
How would that look? It would be much better to not have to pollute the patch with unnecessary config changes.

DemianX0:
I agree that's not very user-friendly. I imagine adding an extra patch to the patchdemo, including only the configuration. That would save the important patch(es) from pollution and only add the price of one extra commit. The benefit: zero complications with another settings file, security, etc.

DemianX0:
@edg2s An example:  https://gerrit.wikimedia.org/r/c/mediawiki/core/+/585856

There are alternative approaches, ex: Patchdemo could generate an optional import at the end of LocalSettings.php for PatchDemoSettings.php

Opinions?

edg2s:
I'd rather we didn't pollute Gerrit with these patches. It's also more of a burden on the developer, they will need to have this cherry pick a config patch to every patch they want to test, and part of the purpose of this tool is to allow non-devs (e.g. QA's) to test patches as well.

DemianX0:
> I'd rather we didn't pollute Gerrit with these patches.
Me too. Until there's a better solution though...

> It's also more of a burden on the developer, they will need to have this cherry pick a config patch to every patch they want to test

No, just apply two patches with patchdemo. See [my last 2 demos](https://patchdemo.wmflabs.org) (list will change with time).

edg2s:
> Me too. Until there's a better solution though...

This task is about coming up with a better solution.

jdlrobson:
Maybe a JSON of config values could be used to generate a LocalSettings.php?

e.g. LocalSettings.json
it would avoid arbitary PHP being run at least.

```
{ "Foo": 4, "stuff": [] }
```
would generate
```
$wgFoo = 4;
$wgstuff = [];
```

DemianX0:
Implicit prefixing has many pitfalls, it's safer to use full variable names with "wg" prefix constraint:
```
{ "wgFoo": 4, "wgstuff": [] }
```

[...]

PR #50, opened 2020-04-23
Customise [[Main Page]]

### LINKED ISSUE #7: Customise [[Main Page]] content
It would probably be better if it said something like:

> This wiki was automatically generated by [https://github.com/MatmaRex/patchdemo patchdemo] at `datetime` and applies the following patches:
> * https://gerrit.wikimedia.org/...
> * ...
> 
> This wiki may be deleted in the near future so do not expect links to it to persist.

The main page is setup in `Installer.php#createMainpage`.

edg2s:
Fixes #7

### DIFF
--- a/new.php
+++ b/new.php
@@ -115,7 +115,7 @@ $patchesAppliedText = implode( ' ', $patchesApplied );
 
 $wikiName = "Patch Demo ($patchesAppliedText)";
 
-$mainPage = "This wiki was automatically generated on [$server$serverPath Patch Demo] at ~~~~~ and applies the following patches:\n";
+$mainPage = "This wiki was generated on [$server$serverPath Patch Demo] at ~~~~~ and applies the following patches:\n";
 
 foreach ( $patchesApplied as $patch ) {
 	preg_match( '`([0-9]+),([0-9]+)`', $patch, $matches );
--- a/new.php
+++ b/new.php
@@ -8,6 +8,8 @@ $branch = trim( $_POST['branch'] );
 $patches = $_POST['patches'];
 
 $namePath = md5( $branch . $patches . time() );
+$server = ( isset( $_SERVER['HTTPS'] ) ? 'https://' : 'http://' ) . $_SERVER['HTTP_HOST'];
+$serverPath = preg_replace( '`/[^/]*$`', '', $_SERVER['REQUEST_URI'] );
 
 echo "Your wiki will be available at:";
 echo "<br>";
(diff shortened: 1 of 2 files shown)


DemianX0:
Do you think some lorem ipsum could fit in this patch?
Like http://patchdemo.wmflabs.org/wikis/8db0db8c3dab429e1f76be26fa50ce80/w/index.php/Main_Page

edg2s:
That would be for a very specific purpose which is testing stylesheet patches. There's lots of other feature specific content that would be useful for testing, so it would be better to have a system for preloading bug specific content, similar to #19.

DemianX0:
> it would be better to have a system for preloading bug specific content, similar to #19.

I agree.

new.php
```diff
@@ -113,6 +115,20 @@
 
 $wikiName = "Patch Demo ($patchesAppliedText)";
 
+$mainPage = "This wiki was automatically generated on [$server$serverPath Patch Demo] at ~~~~~ and applies the following patches:\n";
```
MatmaRex, line 118: MediaWiki has magic words for this, you can just write it like below instead of parsing the URLs:
```
[//{{SERVERNAME}} Patch Demo]
```
Or even this, but it seems to always return a HTTP URL on the existing wikis, possibly we should be setting `$wg`-something to override that:
```
[{{SERVER}} Patch Demo]
```

MatmaRex: Ah, I guess we need them anyway…

MatmaRex:
I tested this and the two edits came out in the wrong order :o http://patchdemo.wmflabs.org/wikis/cf15efa04d9f2274823035fecc005bd3/w/index.php?title=Main_Page&action=history

MatmaRex:
Can we just add a one second sleep to avoid this? I can't be bothered figuring out how to do it better.

edg2s:
That's very strange. Is install.php not synchronous?

What if you add `sleep 1` or `sleep 2` between install and edit?

edg2s:
When I first tested locally, the edit has a timestamp 2 seconds older than the initial commit. On the wiki you showed, they have identical timestamps (to the second): http://patchdemo.wmflabs.org/wikis/cf15efa04d9f2274823035fecc005bd3/w/index.php/Special:ApiSandbox#action=query&format=json&prop=revisions&titles=Main%20Page&rvlimit=max

I suspect sleeping for 2 seconds would work?

MatmaRex:
I think it is synchronous, and the two edits have revision IDs 1 and 2 in the expected order, but apparently the history view only orders by timestamp, and they were the same here.

edg2s:
Ok, so it's just the identical timestamps causing confusion? I think `sleep 1` should be sufficient then.

edg2s:
@Tchanders points out that this is due to an old limitation in MediaWiki pagers only allowing you to sort by one field, so timestamp was chosen as revision id is less reliable. Will link to an upstream task about this later, but another interesting side effect she points out is that when these two edits span a results page, one of them goes missing:
If you login as Patch Demo, I've set preferences to only show 1 result per page. Go to this page: http://patchdemo.wmflabs.org/wikis/cf15efa04d9f2274823035fecc005bd3/w/index.php?title=Main_Page&action=history try to view the second page of results ("older 1"), and you get nothing!

Tchanders:
Here's the task FYI: https://phabricator.wikimedia.org/T251044

new.php
DannyS712, line 118: `applies` should be `applied` for consistency with `was automatically generated`
But, the wiki wasn't "automatically" generated - a user had to choose to created it, so perhaps `was generated`?

edg2s: That gets you to the root, but not if you have a path, e.g. myserver.com/patchdemo/ which currently works everywhere else.

edg2s: Removed "automatically". I'm using "applies" in the continuous tense in that the server makes use of (and continues to make use of) the patches.

Issue #51, opened 2020-04-24
Loosen jenkins-bot approval requirements


DemianX0:
I was trying to create a demo for  https://gerrit.wikimedia.org/r/c/mediawiki/skins/Vector/+/592325

The [previous patch](https://gerrit.wikimedia.org/r/c/mediawiki/skins/Vector/+/584694/25) in chain fails the build, I can't do anything about that.

Demo creation failed because of this:

> Querying patch metadata...
> https://gerrit.wikimedia.org/r/changes/?q=change:592325&o=LABELS&o=CURRENT_REVISION
> https://gerrit.wikimedia.org/r/changes/mediawiki%2Fskins%2FVector~master~I182eedeff0b6eb62250f2682b89f30c8ff0f70de/revisions/ae3b99d9a3c4a1708936eda50ef883e4eb3c3c60/related
> https://gerrit.wikimedia.org/r/changes/592325/revisions/5/commit
> https://gerrit.wikimedia.org/r/changes/584694/revisions/25/commit
> https://gerrit.wikimedia.org/r/changes/?q=change:I8e153c0ab927f9d880a68fb9efb0bf37b91d26b2&o=LABELS&o=CURRENT_REVISION
> Patch must be approved (Verified+2) by jenkins-bot

I'm solving this by squashing the two patches, however that's not optimal as now the patch is not useful to present the changes I've applied.

Optional: Futhermore, it would be helpful to have an option to create a demo with an unapproved patch. jenkinsbot does certain strict tests that are necessary to pass for live code, but not for demonstration code. Sometimes it's not worth the time it takes to investigate and fix every issue jenkins reports.

edg2s:
This feature already exists but is disabled on the wmflabs install for security. Personally I don't think it adds much security as you can still maliciously or accidentally deploy an insecure or dangerous environment.

DemianX0:
> This feature already exists but is disabled on the wmflabs install for security. 

I assume you mean completely disabling the jenkins requirement?
It would be great to have the first part: only requiring jenkins approval for the patch applied, not its ancestors (imo that's more like a fix than a feature) and I have a feeling https://github.com/MatmaRex/patchdemo/issues/52#issuecomment-618924922 would accomplish this too.

MatmaRex:
[...]

PR #60, opened 2020-05-09
Allow site config to be overridden by JSON

### LINKED ISSUE #19: Provide a way to override configs

edg2s:
Fixes #19

### DIFF
--- a/LocalSettings.txt
+++ b/LocalSettings.txt
@@ -63,3 +63,16 @@ $wgGENewcomerTasksRemoteApiUrl = 'https://en.wikipedia.org/w/api.php';
 $wgGENewcomerTasksTopicType = 'ores';
 $wgWelcomeSurveyExperimentalGroups['exp2_target_specialpage']['range'] = '0-9';
 $wgGEHomepageMentorsList = 'Project:GrowthExperiments_mentors';
+
+// Apply config.json
+$config = json_decode(
+	file_get_contents( 'config.json' ),
+	true
+);
+if ( $config ) {
+	foreach ( $config as $key => $value ) {
+		$name = "wg$key";
+		$$name = $value;
+	}
+}
+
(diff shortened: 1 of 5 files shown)


MatmaRex:
I'm 99% sure this allows you to execute arbitrary PHP, just win some extra steps.

There are various config variables that let you read from (e.g. `$wgExtensionMessagesFiles`) or write to (`$wgDebugLogFile`) arbitrary files. Or maybe just add "eval" to `$wgHooks` for some hook that handles user input. I haven't figured out how exactly to exploit it, but it almost certainly can be done.

I think we should figure out a different way of whitelisting (https://github.com/MatmaRex/patchdemo/issues/51#issuecomment-619404434) or something.

edg2s:
> I'm 99% sure this allows you to execute arbitrary PHP, just win some extra steps.

As a V+1 user I can create a patch that does the same thing, so surely this is no less secure?

MatmaRex:
Yes but we won't demo your patch unless it has V+2.

edg2s:
I mean V+2 - you can write a pretty malicious patch and still get V+2 as long as you are trusted. If you aren't trusted you can't create a patch demo anyway.

DemianX0:
1. Am I correct to assume as a non-trustedcontributor one can still create a demo from somebody's V+2 patch with arbitrary config.json, that could be malicious? 

2. What's the abuse potential? I guess dumping these wikis, gaining access to passwords preferably not used anywhere.

edg2s:
> I think we should figure out a different way of whitelisting (#51 (comment)) or something.

The problem with the email whitelist is that OAuth doesn't give us the user email, so we would need to create a separate SUL-based account whitelist.

edg2s:
See #112

PR #82, opened 2020-07-13
LS: Update Parsoid config to new system

jdforrester:
This maybe should work?

### DIFF
--- a/LocalSettings.txt
+++ b/LocalSettings.txt
@@ -13,19 +13,12 @@ $wgDebugComments = true;
 $wgResourceLoaderStorageEnabled = false;
 
 // Parsoid/PHP required configuration
-$wgEnableRestAPI = true;
-wfLoadExtension( 'Parsoid', 'parsoid/extension.json' );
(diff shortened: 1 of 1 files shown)


MatmaRex:
I haven't merged this yet because I want Patchdemo to work with older release branches too and not just master (note the "Start with version" dropdown). Admittedly I haven't tested that since the first commits in this repo, so it might be broken right now, but I don't want to make it impossible to fix. I need to think about it (or you can think about it if you want and propose a solution ;) ).

edg2s:
Speaking of "start with version": #89

MatmaRex:
Looking at this again: we definitely need to keep at least the `wfLoadExtension( … )` call, to load Parsoid from the Git repo (where we can apply the desired patches) rather than from the Composer library (where we can't). I'm not sure if the config options will magically configure themselves right if we do that, or if we need to keep some of them too.

MatmaRex:
I want Patchdemo to stay compatible at least with the non-EOL MW release versions.

PR #85, opened 2020-07-30
Add WikiLambda

jdforrester:


edg2s:
I assume this won't change environments in a way that makes them less production-like, so fine with me.

PR #94, opened 2020-08-07
Build form controls with OOUI

edg2s:


includes.php
```diff
@@ -14,12 +14,15 @@
...
 
+OOUI\Theme::setSingleton( new OOUI\WikimediaUITheme() );
+
 echo '<!DOCTYPE html>
 <html>
 	<head>
 		<meta charset="utf-8">
 		<title>Patch demo</title>
 		<link rel="stylesheet" href="index.css">
+		<link rel="stylesheet" href="node_modules/oojs-ui/dist/oojs-ui-wikimediaui.min.css">
```
edg2s, line 25: we don't deploy with `npm i` at the moment, so perhaps this file should just be copied to a lib folder?

MatmaRex:
This is another dependency that I don't really want to have and that someone will have to update regularly until the end of times. Eh…

includes.php
edg2s: Adding `npm install --production` to the deploy step is probably the easiest.

MatmaRex:
Alright, fine, let's do this… Thank you for updating the patch.

I have some tweaks I want to make, I'll push them here for you to have a look.

MatmaRex:
I think this should work, but apparently the `<summary>` tag stops being clickable when you put a `<label>` inside of it. That's a bit dumb.

I need to update the JS code for disabling checkboxes too.

I'll finish this tomorrow, bleh.

MatmaRex:
Never mind, I finished it today. I think I like it, please have a look at my changes.

MatmaRex (approved):

edg2s:
The span inside details is display:block:
image

edg2s:
Fixed that, LGTM now.

MatmaRex:
Thanks, I only tested on Firefox and I didn't see that.

Some screenshots for future reference:

Firefox:
| Before | After |
| -- | --|
|  | 
|  | 

Chromium:
| Before | After |
| -- | --|
|  | 
|  |

PR #95, opened 2020-08-12
Add GrowthExperiments (and PageViewInfo, which it depends on)

catrope:
Making all the features work required a little bit of configuration and
a few new pages, but it wasn't too bad.

Also add a pages/ directory for static wiki pages to add on each wiki.
The Main Page can't be in here, because it's dynamically generated.

The MediaWiki:NewcomerTasks.json and MediaWiki:NewcomerTopicsOres.json
pages shouldn't be needed in theory, because GrowthExperiments
supports remote pages hosted on mediawiki.org, but that code is buggy
when the URL structures of the source and target wikis differ, as they
do here.


pages/Talk:Main_Page
```diff
@@ -0,0 +1 @@
+Placeholder discussion page
```
edg2s, line 1: What is this for?

LocalSettings.txt
```diff
@@ -46,3 +46,13 @@ $wgLogos = [
...
+
+// PageViewInfo
+$wgPageViewInfoWikimediaDomain = 'en.wikipedia.org';
+
+// GrowthExperiments
+$wgGERestbaseUrl = 'https://en.wikipedia.org/api/rest_v1';
+$wgGENewcomerTasksRemoteApiUrl = 'https://en.wikipedia.org/w/api.php';
+$wgGENewcomerTasksTopicType = 'ores';
+$wgWelcomeSurveyExperimentalGroups['exp2_target_specialpage']['range'] = '0-9';
+$wgGEHomepageMentorsList = 'Mentors';
```
edg2s, line 58: I would make this page name more specific so it's less likely to clash with imported data, something like "Project:GrowthExperiments/Mentors"

edg2s:
Pages which are essentially extension config seem fine, but we may want to think carefully about how we allow test content, as that has the potential to make the wikis very messy.

pages/Talk:Main_Page
catrope: I forgot to comment on this: this is needed to make the help desk features work, because they need a talk page to post to, and the default setting for this is Talk:Main_Page. I'll add a sentence to this page explaining that.

LocalSettings.txt
catrope: I initially put it in the Project namespace, but that leads to an incredibly ugly (normalized) page title like "Patch_Demo_(616810,6):GrowthExperiments/Mentors". You're right that it does belong in the Project namespace though, so I'll put it there.

catrope:
> Pages which are essentially extension config seem fine and seem generally helpful (e.g. Citoid configuration), but we may want to think carefully about how we allow test content, as that has the potential to make the wikis very messy.

I should also clarify (and will update the commit message for this) that I initially planned to use GE's remote config page feature to use config pages hosted on mw.org (which is what most of us do for our local installs), but this doesn't work because of bugs in GE (it assumes the URL structure of the two wikis is identical, and that leads it to request `https://www.mediawiki.org/patchdemo/wikis/123abc/w/index.php?....`)

LocalSettings.txt
edg2s: I think setting the canonical project namespace to "Project" would be sensible.

catrope: Submitted as a separate PR: #96

MatmaRex:
The colons in the file names like "Project:GrowthExperiments_mentors" are going to be particularly annoying for me, as that is not a valid file name on Windows ;)

I can probably deal with that, it's a neat way to define the pages otherwise.

MatmaRex:
The changes in createwiki.sh should be wrapped in an `if`, like in https://github.com/MatmaRex/patchdemo/pull/98/files#diff-bbd4f3310bfa4731632e1a2a1956711b (I just added a way to disable individual extensions).

edg2s:
We could replace ':' with '|' which will never appear in a title. Is that allowed on Windows?

MatmaRex:
No, there's like 8 different invalid characters.

But actually, now that I think of it, this is also going to have a problem if we ever want to define a page with a name containing `/`. There should probably be a better way to define the pages, but we don't need to invent it right now.

MatmaRex:
If anyone is curious, I'm using a sparse checkout locally to avoid dealing with the files with invalid filenames on Windows:
```
git sparse-checkout set /* !pages
```
You need a fairly recent version of Git for this, I've just upgraded to 2.28 to do that.

Also, apparently recent Git will refuse to check out these files; I needed the following to avoid errors like `error: invalid path 'pages/MediaWiki:NewcomerTasks.json'` (even with sparse checkout enabled):
```
git config core.protectNTFS false
```
(This used to not be the case in earlier versions, I was using Git 2.17 until a few hours ago and it happily tried writing the files, resulting in wrong filenames and confusing `git status` output, but not error messages.)

PR #97, opened 2020-08-18
Remove VE wgHiddenPrefs hack now that BetaFeature is hidden by default

edg2s:


edg2s:
Depends on https://gerrit.wikimedia.org/r/c/mediawiki/extensions/VisualEditor/+/620951/ upstream

MatmaRex:
This should probably remain, but only be used for branches older than that commit? I'm not sure how I'd like to do that. For now, it's harmless to keep it…

MatmaRex:
Fair enough.

PR #99, opened 2020-08-18
Sort extensions alphabetically

edg2s:


edg2s:
I notice now that VE core is in the list, which it probably shouldn't be as it doesn't make sense to disable that independently of VE-MW.

MatmaRex:
> I notice now that VE core is in the list, which it probably shouldn't be as it doesn't make sense to disable that independently of VE-MW.

Yeah… But it gets complicated if we start thinking about that, since there are also dependencies between the extensions. For now I prefer to keep the interface "dumb" and expect the user to know what they are doing.

PR #100, opened 2020-08-19
Use correct outline colour for focus

edg2s:


MatmaRex:
What makes it "correct"?

edg2s:
https://phabricator.wikimedia.org/T245887

PR #103, opened 2020-09-02
Fix creating a demo with no patches

MatmaRex:


new.php
```diff
@@ -116,12 +119,24 @@
 	}
 }
 
-$patchesAppliedText = implode( ' ', $patchesApplied );
+$branchDesc = preg_replace( '/^origin\//', '', $branch );
+
+$wikiName = "Patch Demo (" . trim(
+	// Add branch name if it's not master, or if there are no patches
+	( $branchDesc !== 'master' || !$patchesApplied ? $branchDesc : '' ) . ' ' .
```
edg2s, line 126: Put the extra space inside the ternary, otherwise you will get  "( 123456)", and "(REL1.35 )"

edg2s:
I suppose you can now create master+nopatches (for reasons unknown) which will have no title and look a little odd in the table.

new.php
MatmaRex: There's the `trim()` to fix that. (It's annoying to do this without it, because either "side" can be empty.)

MatmaRex:
It will have just "(master)" in the title, but you're right I forgot about the table on the main page.

MatmaRex:
I think that could still be useful for comparing the behavior before/after some patch, so I didn't disallow that.

PR #104, opened 2020-09-02
Support passing no patches

### LINKED ISSUE #102: Should be possible to create a demo with no patches
For example, if I just want to test out an old version of MW.

edg2s:
Fixes #102


edg2s:
Will go with #103

MatmaRex:
Aaahh, sorry

PR #107, opened 2020-09-03
Add custom 404 handler

### LINKED ISSUE #106: Add a custom 404 handler
As we are continually posting links to wikis and then deleting them, we should probably have a 404 handler that explains this.

edg2s:
Fixes #106

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,5 @@
+RewriteEngine on
+# Equivalent to ErrorDocument, but uses a relative path
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteCond %{REQUEST_FILENAME} !-d
+RewriteRule .* 404.php [L]
--- a/404.php
+++ b/404.php
@@ -1,0 +1,10 @@
+<?php
+header( 'HTTP/1.0 404 Not Found' );
+require_once "includes.php";
+
+echo new \OOUI\MessageWidget( [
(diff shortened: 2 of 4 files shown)


edg2s:
Merge after #94.

MatmaRex:
The paths don't work on https://patchdemo.wmflabs.org/, because `dirname( '/index.php' )` returns a slash rather than an empty string, I think this just needs to be special-cased. It probably works fine when Patchdemo is in a subdirectory.

Also I don't think `$IP` is the right name, in MediaWiki that refers to the filesystem path (i.e. /var/www/html/index.php), rather than the URL path (i.e. /index.php). The MediaWiki equivalent is `$wgScriptPath`, but I'd just call it the base path or something.

MatmaRex:
I'm now reading up on how to enable .htaccess files, and how to do that without manually editing weird config files, so that I can put it in setup.sh…

edg2s:
You need to modify the `AllowOverride` config. I'm pretty sure you need access to the core Apache config (i.e. be the root user) because it's a security feature.

edg2s:
If the user hasn't set it up it isn't the end of the world, it's a progressive enhancement.

MatmaRex:
Yeah but I'm not willing to figure that out again when something inevitably hoses the site one day, so I'm going to automate it. #108

PR #108, opened 2020-09-08
Enable .htaccess files and mod_rewrite

MatmaRex:


MatmaRex:
I think this is the "proper" way (that won't cause trouble when upgrading Apache and it comes with new defaults).

edg2s (approved):
It makes a few assumptions (e.g. files being in /var/www/html) but I guess there is other stuff in setup.sh that might need to be tweaked.

PR #109, opened 2020-09-17
Query patch metadata first when creating

edg2s:
The mostly likely failure scenario is that the user inputted patches that can't be merged, so do this first so the error (1) displays sooner and (2) is shown at the top of the page.


MatmaRex:
(rebased)

PR #110, opened 2020-09-17
Don't share cookies across wikis

### LINKED ISSUE #78: Session cookies piling up
> Bad Request
Your browser sent a request that this server could not understand.
Size of a request header field exceeds server limit.
Apache/2.4.38 (Debian) Server at patchdemo.visualeditor.eqiad.wmflabs Port 80

Happened on https://patchdemo.wmflabs.org/ after oauthing.
It turns out, session cookies have `Path == '/'` set.

I've managed to hoard a considerable stash, enough for a Christmas party, sent around with each request. If only we could eat them...

Can you limit the session cookies to the path of the individual wikis?

edg2s:
Fixes #78

### DIFF
--- a/LocalSettings.txt
+++ b/LocalSettings.txt
@@ -8,6 +8,9 @@ $wgShowDBErrorBacktrace = true;
 $wgShowExceptionDetails = true;
 $wgDebugComments = true;
 
+// Don't share cookies across wikis (#78)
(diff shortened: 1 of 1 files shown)


MatmaRex:
If we ever set up proper short URLs for the wikis, $wgCookiePath will need to be updated to not include the `/w` path.

PR #112, opened 2020-09-17
Allow site config to be overridden by trusted users

### LINKED ISSUE #19: Provide a way to override configs

edg2s:
Fixes #19

### DIFF
--- a/config.default.php
+++ b/config.default.php
@@ -2,6 +2,8 @@
 $config = [
 	// Allow any user to delete wikis, e.g. on a private installation
 	'allowDelete' => false,
+	// Allow any user to add site config, e.g. on a private installation
+	'allowSiteConfig' => false,
 	// Require that patches are V+2 before building the wiki
 	'requireVerified' => true,
 	// OAuth config. When enabled only authenticated users can create
@@ -12,6 +14,11 @@ $config = [
 		'key' => null,
 		'secret' => null,
 		// OAuth admins can delete any wiki
-		'admins' => []
+		'admins' => [],
+		// These users can override site configs. This is the same level of trust as V+2,
+		// as those users can also execute arbitrary code.
+		'configurers' => [],
+		// Same as above, but regexes e.g. / \(WMF\)$/
+		'configurersMatch' => [],
 	]
 ];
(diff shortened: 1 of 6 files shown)


edg2s:
This does away with the JSON approach of #60 as that adds no real JSON.

Instead we can maintain a list of trusted OAuth accounts that will be allowed to use the site config feature.

To begin with we should use the regex feature to allow `/ \(WMF\)$/` as I believe these names are protected.

DannyS712:
Would love to be able to do this, but how exactly does the code detect if a user is V+2? OAuth connects to SUL, and being able to vote V+2 is based on gerrit permissions on an account that may not be connected.

edg2s:
This patch requires us to manually maintain a list of trusted users.

MatmaRex:
I like this but I'd also like to avoid maintaining that list for the rest of my life, hmm…

edg2s:
One could host the user list in a Gerrit repo, then v+2ers could add themselves.

DannyS712:
> One could host the user list in a Gerrit repo, then v+2ers could add themselves.

That would also require CR+2, though that should not be a problem since the two usually go together.
That would also ensure that the list only includes those that want to be included and plan to use the tool

Sounds like a good idea to me. New repo, something like `mediawiki/tools/patchdemoconfig` (in case the actual code from this ever moves, not using `patchdemo`) inheriting from `mediawiki/*` with a simple text file that users are told they can self-merge changes to, file just holds a list of allowed user names

config.default.php
```diff
@@ -14,7 +14,14 @@
...
 		// OAuth admins can delete any wiki
-		'admins' => []
+		'admins' => [],
+		// These users can override site configs. This is the same level of trust as V+2,
+		// as those users can also execute arbitrary code.
+		'configurers' => [],
+		// Same as above, but regexes e.g. / \(WMF\)$/
+		'configurersMatch' => [],
+		// Instructions to request 'configurers' user status, e.g. "File a request <a href=...>here</a>."
+		'configurersRequestHtml' => '',
```
edg2s, line 24: Suggested config for us:
`Request approval by creating a <a href="https://github.com/MatmaRex/patchdemo/issues/new">new issue</a>.`

PR #121, opened 2020-09-22
Add space for a banner message

MatmaRex:


includes.php
```diff
@@ -154,6 +154,11 @@ function user_link( $username ) {
 	return '<a href="' . $base . '?title=' . urlencode( 'User:' . $username ) . '" target="_blank">' . $username . '</a>';
 }
 
+function banner_html() {
+	global $config;
+	return $config['banner'] ?? '';
```
edg2s, line 159: configs are extended from default, so we can assume 'banner' exists

index.php
```diff
@@ -95,6 +95,11 @@
 		] ),
 	]
 ] );
+
+$banner = banner_html();
+if ( $banner ) {
+	echo "<p class='banner'>$banner</p>";
```
edg2s, line 101: Should we use a MessageWidget with type=warning (or info)?

edg2s (changes requested):
.

index.php
edg2s: Also, should the banner go above the form, given that it should be read before creating a wiki?

MatmaRex: I don't want to put it at the top because I think no one will read it then :D I was considering putting it in the middle of the form though (above the button), but I like it here.

I greatly dislike how MessageWidget looks.

edg2s:
Do you have text for the banner so I can test?

MatmaRex:
I updated the config on https://patchdemo.wmflabs.org/

edg2s:
Screenshot:
image

edg2s:
I think an alert icon would draw your attention a bit more, this could be done by setting the banner as a warning on the submit field:

image

MatmaRex:
Eeeehh

edg2s:
Word-smithing:

"Any wiki created with this tool may be deleted at any time along with all the content on it, so you should not rely on them being available, or use them to store important information."

I don't think we want to say "_for anything_" as you can rely on the wikis to check out your patch correctly, for example.

index.php
edg2s: We should call this "[new/create]Wiki[Banner/Warning]", as "banner" suggests is appears everywhere.

If we add other banners later (e.g. on new.php, or Main_Page) that's one less thing to change in your config.

MatmaRex:
Sure, thanks.

>you can rely on the wikis to check out your patch correctly, for example

"rely" is such a strong word…

PR #123, opened 2020-09-25
Convert banner to field warning

edg2s:


edg2s:
Before:
image
After:
image

MatmaRex:
I know it's a small change, but I really hate it.

edg2s:
> I know it's a small change, but I really hate it.

Tell me what you really think though.

edg2s:
What about an info icon? I think it should probably match the width of the form either way?

MatmaRex:
Really please no

If it should match the width of the form, then I think the form should be allowed to be wider. I'm actually not sure why its width was limited, when the width of the whole page is also already limited? I feel like it would be fine if it was expanded from `50em` to `1100px`, like the table and everything else.

edg2s:
> If it should match the width of the form, then I think the form should be allowed to be wider. I'm actually not sure why its width was limited, 

Forms becomes harder to use when you make them too wide. Either the labels are too far from the inputs, or the inputs are too wide and the select handle is too far from the labels.

index.css
```diff
@@ -86,6 +86,22 @@ form {
 	max-width: 50em;
 }
 
+/* T263737 */
```
MatmaRex, line 89: No longer needed?

MatmaRex (changes requested):

MatmaRex:
I'm still not convinced that this is an improvement.

MatmaRex:
Ah, whatever.

PR #124, opened 2020-09-27
Add GlobalWatchlist and GuidedTour extensions

DannyS712:
GlobalWatchlist extension is being developed for use on WMF wikis, and it uses the GuidedTour extension when available


MatmaRex:
Does GlobalWatchlist do anything here? The patchdemo wikis are not a wiki farm or connected in any way.

DannyS712:
Yes, it works on a single wiki as well

PR #127, opened 2020-09-29
Adding FAQ

PTPells:
Adding the questions I've/we've received from WMF staff. Two of the questions I didn't know the answers to. They are marked as `#TODO`

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ FAQ
 ----
 **Can you delete it when you are done with it?**
 
-Yes. For any wiki you create, you will see a `Delete` link in the `Action` column of this table: https://patchdemo.wmflabs.org/
+Yes. For any wiki you create, you will see a `Delete` link in the `Action` column of this table: https://patchdemo.wmflabs.org/. We advise you to delete the wikis you create when you are finished with them and/or when the patch you created the wiki to test is merged.
 
 **How long do the Patch demo wiki instances last for?**
 
@@ -35,12 +35,18 @@ There is no definitive time after which wikis will automatically be deleted. Wit
 
 **Can Patch demo wikis be named?**
 
-Wikis can not been named *within* Patch demo. Patch demo wikis currently inherit the name of the patch on which they are built.
+Wikis can not been named *within* Patch demo. Wikis are listed within Patch demo by the creator and the list of patches (potentially multiple) used to create it. They are also assigned a random hash, which becomes part of the URL. 
 
 **Is it possible to add extensions that are in development?**
 
-#TODO
+These will be considered on a case-by-case basis, but will generally be allowed as long as they don't interfere with other teams' ability to test in a production-like environment.
 
 **Is it possible to add patches for extension not just core? And skins?**
 
-#TODO
+Yes, patches for many extensions and skins are supported (mostly those included in MediaWiki releases, or enabled on all Wikimedia wikis), as well as Parsoid. Check out the list under "Choose extensions to enable" in the interface.
+
+**What happens to a Patch demo wiki when the underlying patch is updated?**
+
+Nothing. Once created, the wikis are never updated. New versions of the selected patches are not applied, and neither are patches merged into master. If you want to test a newer version of the patch, create a new wiki with it.
+
+
--- a/README.md
+++ b/README.md
@@ -22,3 +22,25 @@ Tested on Ubuntu 18.04 and Debian 10.
 * Put all these files in `/var/www/html`
 * Run `sudo setup.sh`
 * Visit http://localhost in your browser
+
+FAQ
+----
+**Can you delete it when you are done with it?**
+
+Yes. For any wiki you create, you will see a `Delete` link in the `Action` column of this table: https://patchdemo.wmflabs.org/
+
+**How long do the Patch demo wiki instances last for?**
+
+There is no definitive time after which wikis will automatically be deleted. With this said, we make no guarantees about how long they will continue to exist. A Patch demo wiki you've created could be deleted if we need to free up disk space to create space for new ones.
+
+**Can Patch demo wikis be named?**
+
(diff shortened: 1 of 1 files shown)


PTPells:
@MatmaRex another question to add:

**What happens to a Patch demo wiki when the underlying patch is updated?**

README.md
```diff
@@ -22,3 +22,25 @@ Tested on Ubuntu 18.04 and Debian 10.
 * Put all these files in `/var/www/html`
 * Run `sudo setup.sh`
 * Visit http://localhost in your browser
+
+FAQ
+----
+**Can you delete it when you are done with it?**
+
+Yes. For any wiki you create, you will see a `Delete` link in the `Action` column of this table: https://patchdemo.wmflabs.org/
```
edg2s, line 30: I would add that people **should** delete wikis when they are done, or the patch is merged.

README.md
```diff
@@ -22,3 +22,25 @@ Tested on Ubuntu 18.04 and Debian 10.
...
+
+**How long do the Patch demo wiki instances last for?**
+
+There is no definitive time after which wikis will automatically be deleted. With this said, we make no guarantees about how long they will continue to exist. A Patch demo wiki you've created could be deleted if we need to free up disk space to create space for new ones.
+
+**Can Patch demo wikis be named?**
+
+Wikis can not been named *within* Patch demo. Patch demo wikis currently inherit the name of the patch on which they are built.
```
edg2s, line 38: It depends what you mean by "name". There is no "name" label in the table. Wikis are listed by creator and the list of patches (potentially multiple) used to create it. They are also assigned a random hash, which becomes part of the URL. If you are referring to $wgSitename, which is used to set the page title and other labels on the wiki itself, then that can not currently be set, although it may be settable in the future.

README.md
```diff
@@ -22,3 +22,25 @@ Tested on Ubuntu 18.04 and Debian 10.
...
+
+FAQ
+----
+**Can you delete it when you are done with it?**
+
+Yes. For any wiki you create, you will see a `Delete` link in the `Action` column of this table: https://patchdemo.wmflabs.org/
+
+**How long do the Patch demo wiki instances last for?**
+
+There is no definitive time after which wikis will automatically be deleted. With this said, we make no guarantees about how long they will continue to exist. A Patch demo wiki you've created could be deleted if we need to free up disk space to create space for new ones.
```
edg2s, line 34: "Wikis are not currently automatically deleted"

README.md
```diff
@@ -22,3 +22,25 @@ Tested on Ubuntu 18.04 and Debian 10.
...
+**Can Patch demo wikis be named?**
+
+Wikis can not been named *within* Patch demo. Patch demo wikis currently inherit the name of the patch on which they are built.
+
+**Is it possible to add extensions that are in development?**
+
+#TODO
+
+**Is it possible to add patches for extension not just core? And skins?**
+
```
edg2s, line 45: Yes, any extension or skin that is currently installed (see repositories.txt) can have a patch applied.

README.md
```diff
@@ -22,3 +22,25 @@ Tested on Ubuntu 18.04 and Debian 10.
...
+**How long do the Patch demo wiki instances last for?**
+
+There is no definitive time after which wikis will automatically be deleted. With this said, we make no guarantees about how long they will continue to exist. A Patch demo wiki you've created could be deleted if we need to free up disk space to create space for new ones.
+
+**Can Patch demo wikis be named?**
+
+Wikis can not been named *within* Patch demo. Patch demo wikis currently inherit the name of the patch on which they are built.
+
+**Is it possible to add extensions that are in development?**
+
```
edg2s, line 41: These will be considered on a case-by-case basis, but will generally be allowed as long as they don't interfere with other teams' ability to test in a production-like environment.

README.md
MatmaRex: (I don't want to allow setting names because I don't want people to start relying on a particular wiki existing for a long time.)

README.md
MatmaRex: I would phrase it along the lines of:

Yes, patches for many extensions and skins are supported (mostly those included in MediaWiki releases, or enabled on all Wikimedia wikis), as well as Parsoid. Check out the list under "Choose extensions to enable" in the interface.

MatmaRex:
> What happens to a Patch demo wiki when the underlying patch is updated?

Nothing. Once created, the wikis are never updated. New versions of the selected patches are not applied, and neither are patches merged into master. If you want to test a newer version of the patch, create a new wiki with it.

MatmaRex:
(Thanks for writing these down)

README.md
PTPells: Done.

README.md
PTPells: Sounds good. I've updated the PR to include both points above.

README.md
PTPells: Done.

README.md
PTPells: Done.

edg2s:
Maybe we should move these FAQ to mediawiki.org? We could link to said page in our banner message (the one that starts "Any wiki created with this tool...").

I say this because strictly speaking there is a distinction between our installation on patchdemo.wmflabs.org, and the software in general which someone could install locally. Some of these FAQ are specific to the patchdemo.wmflabs.org instance.

MatmaRex:
Thanks @PTPells

MatmaRex:
> Maybe we should move these FAQ to mediawiki.org? We could link to said page in our banner message (the one that starts "Any wiki created with this tool...").
> 
> I say this because strictly speaking there is a distinction between our installation on patchdemo.wmflabs.org, and the software in general which someone could install locally. Some of these FAQ are specific to the patchdemo.wmflabs.org instance.

I'm trying to clarify this in #142

PR #128, opened 2020-09-29
Enable Citoid

edg2s:
* Add required JSON config files
* Add 4 citation templates with real TemplateData (from en.wiki) and simplified implementations.


edg2s:
Making the citation templates match live exactly would require importing dozens of modules and templates. With this config and some simplified templates we can test the VE-Citoid workflow, even if the rendering isn't quite perfect:
image

PR #130, opened 2020-09-29
Display the table of previously generated wikis when not signed in

MatmaRex:
The diffs are huge because I wrapped a bunch of code in big if()-s, ignore whitespace when reviewing.


MatmaRex:
https://github.com/MatmaRex/patchdemo/pull/130/files?w=1

edg2s (approved):

edg2s:
I guess the only downside here is that the wikis become more crawl-able and possibly more subject to spam. If this becomes a problem we should maybe reconsider this.

PR #134, opened 2020-09-30
Use XML dumps for importing default content

### LINKED ISSUE #129: Don't store demo page titles as filename
As discussed when this feature was merged (#95) there are problems with storing the page title as the filename:
* Windows doesn't like colons `:` and probably other punctuation that is valid in a page title
* All OSes won't allow `/` in a filename, i.e. subpages

Other options:
* Use a descriptive filename, then store the title in the first line of the file
  * Easy to read/create
  * Requires a more complex script to import
* Use the [[Special:Export]] XML format
  * Can be imported using existing scripts
  * Can store author/page history if required

cc @catrope @MatmaRex

edg2s:
Also removes check for GrowthExperiments extension before importing content.

Fixes #129


edg2s:
CC @catrope

edg2s:
I generated the GrowthExperiments.xml dump by Special:Export-ing the following files from a wiki created with patchdemo before this patch:
```
MediaWiki:NewcomerTasks.json
MediaWiki:NewcomerTopicsOres.json
Project:GrowthExperiments_mentors
Talk:Main_Page
```

Citoid.xml was generated the same way and I've tested that locally.

MatmaRex:
As an interesting side-effect, this results in those pages being created with edit dates older than when the wiki was created (whatever is written in the XML file).

I'm not sure how the import works, maybe we can omit some of the metadata fields. I don't really care enough to experiment with it, though.

edg2s:
Yeah, maybe there's an option in the import script. Doesn't really matter though...

PR #135, opened 2020-10-01
Tell users when they have wikis containing only merged patches

### LINKED ISSUE #133: Tell users when they have wikis containing only merged patches
Along the theme of cleaning up old wikis (#14):

We can show a reminder along the lines of "7 of your wikis were built from patches than have now been merged. You should consider deleting these to free up space."

edg2s:
- De-indent table script block
- Add checkbox to filter only "closed" wikis
- Prompt users when they have closed wikis than can be deleted

Fixes #133

### DIFF
--- a/index.css
+++ b/index.css
@@ -115,6 +115,10 @@ summary > .oo-ui-labelElement-label {
 	display: none;
 }
 
+.wikis.hideOpen tr.open {
+	display: none;
+}
+
 .wikis th,
 .wikis td {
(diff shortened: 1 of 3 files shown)


MatmaRex:
I'd prefer only doing this for merged, not abandoned

(also, needs rebase)

edg2s:
It's still up the the user on a case by case basis, but more often than not I will have given up an abandoned patch permanently, so it no longer has testing value.

PR #138, opened 2020-10-01
Add a bit of formatting to the generated Main_Page

edg2s:


edg2s:
Before:
image

After:
image

PR #141, opened 2020-10-02
Enable uploads

edg2s:
Required for testing UploadWizard (and plain uploads)


MatmaRex:
Hmm, you'll probably also need to make some tweaks to php.ini about the max upload/post size, IIRC the defaults are super tiny and nothing works.

edg2s:
It's only for testing...

PR #148, opened 2020-10-13
Use button styling for the button to show closed wikis

MatmaRex:


edg2s:
I've never been a fan of buttons in line with text, maybe because the alignment never seems quite perfect.

_Before_
image

_After_
image

But if you prefer this, go for it. Code looks fine.

edg2s:
I think this would look better as a warning box, with a right-aligned button:

image

cf GitHub notifications:
image

MatmaRex:
Eh, good enough.

MatmaRex:
(Your proposal also looks okay)

PR #149, opened 2020-10-14
Set session expiration to a month

### LINKED ISSUE #75: Store OAuth credentials so session doesn't expire after 24 minutes

MatmaRex:
Fixes #75

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -28,6 +28,10 @@ GRANT ALL PRIVILEGES ON \`patchdemo\_%\`.* TO 'patchdemo'@'localhost';
 composer update --no-dev
 npm install --production
 
+# set session expiration to a month (default is 24 minutes???), cookie expiration too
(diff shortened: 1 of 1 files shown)


edg2s:
I just stayed logged in for a few hours.

setup.sh
```diff
@@ -28,6 +28,10 @@ GRANT ALL PRIVILEGES ON \`patchdemo\_%\`.* TO 'patchdemo'@'localhost';
 composer update --no-dev
 npm install --production
 
+# set session expiration to a month (default is 24 minutes???), cookie expiration too
```
DemianX0, line 31: This sounds like 24*60 instead of 24*60*60: in good PHP manners someone forgot to convert minutes to seconds.
And indeed it is: https://stackoverflow.com/questions/14703363/why-is-the-standard-session-lifetime-24-minutes-1440-seconds/37176824#37176824

PR #155, opened 2020-10-18
Add repository presets and interface to pick them

### LINKED ISSUE #125: Make it easier to only enable a few extensions
Is it possible to have a button to invert the currently selected extensions/skins to be enabled, or to set them all to disabled, rather than needing to click each manually? (Though it should probably require a skin be enabled so that it works)

MatmaRex:
Fixes #125


edg2s:
I like this but I think we need to make it easier to discover which extensions are included or excluded from certain lists. For example if I am a designer on the Growth team, it would be pretty difficult for me to work out that GrowthExperiments is not included in "Wikimedia".

edg2s:
One possibility would be to maintain a separate control to make the list of extensions visible, instead of only showing them when 'Other' is selected. If the user modifiers any of the checkboxes, 'Other' is automatically selected.

edg2s:
We should also notify the teams/people using PD whose extensions will no longer included in the "Wikimedia" default:
* FlaggedRevs (cc @DannyS712)
* GlobalWatchlist (cc @DannyS712)
* GrowthExperiments (cc @catrope)
* ReplaceText (included in the initial commit, why is this in the list?)
* UploadWizard (that was me)
* WikiLambda (cc @jdforrester)

index.php
```diff
@@ -86,15 +88,50 @@
...
+									'data' => 'tarball',
+									'label' => new OOUI\HtmlSnippet( '<abbr title="Skins and extensions included in the official MediaWiki release">Tarball</abbr>' ),
+								],
+								[
+									'data' => 'none',
+									'label' => 'None',
+								],
+								[
+									'data' => 'other',
+									'label' => 'Other',
```
edg2s, line 114: Should this be "Custom"?

MatmaRex:
Tweaked the interface like you suggested.

MatmaRex:
Before | After
------------ | -------------
image | image
image | image

edg2s:
@MatmaRex Do you know why ReplaceText is included?

MatmaRex:
Because it's included in the release tarball (as of 1.35).

PR #160, opened 2020-10-27
Make table headers sticky

edg2s:


MatmaRex:
The `box-shadow` thing is really ugly. I don't think that should ever be used anywhere.

The real problem is that `border-collapse: collapse` and `position: sticky` (or maybe `position` in general) are not compatible – the collapsed borders do not reappear when the cells are being positioned, and they arguably should. But the specs say nothing about this and browsers do not implement any such thing.

So if you really want sticky headers, then we shouldn't be using `border-collapse: collapse`. It's easy enough to do that, just ensure the table cells only have borders on one side, it's just a little tedious.

MatmaRex:
I pushed a patch that does this onto this pull request, GitHub is displaying this in the wrong order…

Please rebase and merge if that looks alright to you (don't squash).

edg2s:
> The box-shadow thing is really ugly.

Actually ugly or code ugly? It looks fine for me. If it's the latter we can fix it later...

PR #162, opened 2020-10-27
Use more OOUI infusion

edg2s:


index.js
```diff
@@ -3,59 +3,44 @@
 	// TODO: Use infuse to control OOUI widgets
 	var myWikis, closedWikis, branchSelect, form, submit, showClosed,
 		presetInput, reposField, reposInput, reposFieldLabel,
-		wikisTable = document.getElementsByClassName( 'wikis' )[ 0 ];
+		$wikisTable = $( '.wikis' );
 
 	function setDisabled( input, disabled ) {
```
edg2s, line 8: Will try and get rid of this later...

index.js
```diff
@@ -76,9 +77,19 @@
...
+			reposInput.checkboxMultiselectWidget.items.forEach( function ( option ) {
+				if ( !option.isDisabled() ) {
+					enabled++;
+					if ( option.isSelected() ) {
+						selected++;
+					}
+				}
+			} );
+
+			reposField.setLabel( reposFieldLabel + ' (' + selected + '/' + enabled + ')' );
```
MatmaRex, line 89: I might not have added this had I realized it needs to be this complex :P

index.js
```diff
@@ -3,75 +3,51 @@
 	// TODO: Use infuse to control OOUI widgets
```
MatmaRex, line 3: Resolved? :)

PR #167, opened 2020-10-28
Run deduplicate.sh as cron job

### LINKED ISSUE #165: rdfind in deduplicate.sh is getting slow, make it a cron job
It takes several minutes to run on the full wikis folder, and all it does is free up some disk space.

edg2s:
Fixes #165

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -11,7 +11,6 @@ while IFS=' ' read -r repo dir; do
 done < ../repositories.txt
 cd ..
 
-
 # Composer wants a directory for itself (COMPOSER_HOME)
 sudo -u www-data mkdir composer
 
@@ -27,6 +26,10 @@ GRANT ALL PRIVILEGES ON \`patchdemo\_%\`.* TO 'patchdemo'@'localhost';
 # dependencies for the website
 composer update --no-dev
 npm install --production
+
+# setup daily cron job to deduplicate files
+BASEDIR=$(dirname "$0")
+sudo ln -s $BASEDIR/deduplicate.sh /etc/cron.daily/
 
 # set session expiration to a month (default is 24 minutes???), cookie expiration too
 echo "session.gc_maxlifetime = 2592000
--- a/deduplicate.sh
+++ b/deduplicate.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-cd $PATCHDEMO
+BASEDIR=$(dirname "$0")
 
 # Using hardlinks assumes that these files will never change,
 # as changing any of the deduplicated copies would affect all of them.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -21,6 +21,7 @@ Tested on Ubuntu 18.04 and Debian 10.
 
 * Put all these files in `/var/www/html`
 * Run `sudo setup.sh`
+* Create a daily cron job as to run `deduplicate.sh` using the same account as your webserver
```
MatmaRex, line 24: The command you added to setup.sh already does this, right? I wouldn't mention it here.

setup.sh
```diff
@@ -28,6 +27,10 @@ GRANT ALL PRIVILEGES ON \`patchdemo\_%\`.* TO 'patchdemo'@'localhost';
 composer update --no-dev
 npm install --production
 
+# setup daily cron job to deduplicate files
+BASEDIR=$(dirname "$0")
+sudo ln -s $BASEDIR/deduplicate.sh /etc/cron.daily/
```
MatmaRex, line 32: I didn't know jobs can be scheduled this way, neat.

edg2s: Neither did I until now.

README.md
edg2s: Oh yeah, I fixed it later

MatmaRex:
I ran the setup script on patchdemo.wmflabs.org, so it should start running daily. But I don't trust cron jobs, so to verify, I just created two identical test wikis and checked their size.
```
matmarex@patchdemo:/var/www/html$ du -sh wikis/c3058b39da82d3c238a3ab3972ded2cc wikis/6f7b4fe31f8353e49cb49f5488cd3807
392M    wikis/c3058b39da82d3c238a3ab3972ded2cc
392M    wikis/6f7b4fe31f8353e49cb49f5488cd3807
```
I am expecting one of them to drop to ~0 after the deduplicate job runs.

MatmaRex:
It does not seem to be working. Those two directories are still not deduplicated.

edg2s:
What if you run the script manually (as root from the root folder)? Does it have execute permissions?

edg2s:
Looks like the symlink is not set up correctly:
image

edg2s:
Let's see if this works tomorrow:
image

edg2s:
I ran `du -sh` on a few wikis created earlier in October before this patch, and they all seem to be ~390M:
image

MatmaRex:
It didn't work still. I'm trying to look into it.

MatmaRex:
The deduplicate script doesn't actually support being called through a symlink. It will try to deduplicate /etc/cron.daily/wikis/ and do nothing. This might help:
```
BASEDIR=$(dirname $(readlink -f "$0"))
```
But more seriously, Cron refuses to run it at all, because of this: https://stackoverflow.com/a/43454485
> the files under these directories have to be pass some sanity checks including the following: be executable, be owned by root, not be writable by group or other and, if symlinks, point to files owned by root. Additionally, the file names must conform to the filename requirements of run-parts: they must be entirely made up of letters, digits and can only contain the special signs underscores ('_') and hyphens ('-'). Any file that does not conform to these requirements will not be executed by run-parts. For example, any file containing dots will be ignored.

PR #168, opened 2020-10-28
Run update script after installation

### LINKED ISSUE #166: Run update.php after installation
In recently created wikis, abusefilter and flaggedrevs tables are not being created.

We can either wait for an upstream fix or just run update.php when creating new wikis.

edg2s:
Fixes #166


MatmaRex:
This should not be needed, but it fixes #166 at the moment, so let's do it I guess.

PR #172, opened 2020-11-02
Allow wikis to be redirected when deleting

### LINKED ISSUE #54: Update wikis with newer patchset, retaining database
Use-case: for some changesets I've set up test users, a demo page. I demoed 4 patches from a chain, a small error in the first patch required a rebase of the whole chain and recreation of the 4 patches, setting up a user + main page on all 4 again. I expect I'll be doing this again and again. An option to update a former wiki would make this a snap.

edg2s:
Fixes #54

### DIFF
--- a/delete.php
+++ b/delete.php
@@ -11,24 +11,88 @@ if ( !can_delete( $creator ) ) {
 }
 
 if ( !isset( $_POST['confirm' ] ) ) {
-	echo '<form method="POST">' .
-		'<p>Are you sure you want to delete this wiki: <a href="wikis/' . $wiki . '/w">' . $wiki . '</a>?</p>' .
-		'<p>This cannot be undone.</p>' .
-		new OOUI\ButtonInputWidget( [
-			'type' => 'submit',
-			'name' => 'confirm',
-			'label' => 'Delete',
-			'flags' => [ 'primary', 'destructive' ]
-		] ) .
-	'</form>';
-	die();
+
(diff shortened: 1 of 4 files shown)


edg2s:
Thoughts?

edg2s:
Thoughts @MatmaRex?

MatmaRex:
I think this could be confusing, e.g. if you're discussing different versions of a patch and then a demo of an older version gets redirected to the newer one.

Maybe we should allow defining a message instead, or at least add an interstitial rather than actually redirect?

edg2s:
Changed to a redirect message instead.

MatmaRex:
If you still want to get this done, I guess it should use the database for storage now…

delete.php
```diff
@@ -11,57 +11,117 @@
...
+					),
+					'items' => array_filter( [
+						count( $wikilist ) > 1 ?
+							new OOUI\FieldLayout(
+								new OOUI\DropdownInputWidget( [
+									'name' => 'redirect',
+									'options' => $wikilist,
+								] ),
+								[
+									'label' => 'Leave a redirect another wiki (optional):',
```
MatmaRex, line 71: Typo: Leave a redirect __to__ another wiki

MatmaRex (changes requested):
(also, needs rebase)

delete.php
edg2s: Fixed

PR #173, opened 2020-11-04
Fix the Cron job to deduplicate files

MatmaRex:
The script was not executed, because it did not pass the "sanity
checks": https://stackoverflow.com/a/43454485

>   the files under these directories have to be pass some sanity checks
>   including the following: be executable, be owned by root, not be
>   writable by group or other and, if symlinks, point to files owned by
>   root. Additionally, the file names must conform to the filename
>   requirements of run-parts: they must be entirely made up of letters,
>   digits and can only contain the special signs underscores ('_') and
>   hyphens ('-'). Any file that does not conform to these requirements
>   will not be executed by run-parts. For example, any file containing
>   dots will be ignored.

Furthermore, because it was a symlink, the actual script would not
correctly detect the directory it was in.

Simplify things by making a regular script that call outs script
instead of a symlink.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -28,8 +28,8 @@ composer update --no-dev
 npm install --production
 
 # setup daily cron job to deduplicate files
-BASEDIR=$(dirname "$0")
-sudo ln -s $BASEDIR/deduplicate.sh /etc/cron.daily/
+echo $(readlink -f deduplicate.sh) > /etc/cron.daily/patchdemo-deduplicate
+chmod u+x /etc/cron.daily/patchdemo-deduplicate
 
 # set session expiration to a month (default is 24 minutes???), cookie expiration too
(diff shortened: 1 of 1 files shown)


MatmaRex:
I deployed this on https://patchdemo.wmflabs.org/, according to `sudo run-parts --test /etc/cron.daily/` it will actually run now, let's see if it will also work as expected.

edg2s:
Neat. If this doesn't work we can just go with my original plan which is to document "Add a cron job" in the README.

edg2s (approved):

MatmaRex:
That should actually work now.

MatmaRex:
Works. https://github.com/MatmaRex/patchdemo/issues/165#issuecomment-723200952

PR #174, opened 2020-11-05
Put wiki link in first column

edg2s:


MatmaRex:
Ah, I've just about gotten used to the silly way we had it…

PR #176, opened 2020-11-10
Revert "Run update script after installation"

MatmaRex:
Reverts #168. The MediaWiki bug https://phabricator.wikimedia.org/T266681 is fixed and this should no longer be needed.


edg2s:
Does this affect any release you can choose from the dropdown?

MatmaRex:
Some of the wmf versions 🤷‍♂️

MatmaRex:
Not worth it

PR #181, opened 2020-12-12
Notify users when wiki is ready

### LINKED ISSUE #178: Notify users when their wiki is ready
Using the notifications API we can tell the user when their wiki has been built. Building a wiki can take a few minutes during which you are likely to switch to another task.

edg2s:
Fixes #178

### DIFF
--- a/css/common.css
+++ b/css/common.css
@@ -98,6 +98,10 @@ summary > .oo-ui-labelElement-label {
 	margin-right: 1em;
 }
 
+.enableNotifications {
+	margin-left: 40%;
+}
+
 /* Wiki list */
 
 .showClosed {
--- a/index.php
+++ b/index.php
@@ -178,6 +178,19 @@ if ( $useOAuth && !$user ) {
 						]
 					),
 					new OOUI\FieldLayout(
+						// Placeholder, will be replaced by a ToggleButtonWidget in JS
+						new OOUI\ButtonWidget( [
+							'icon' => 'bell',
+							'disabled' => 'true'
+						] ),
+						[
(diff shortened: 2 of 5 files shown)


MatmaRex:
I tried this and didn't get a prompt to allow notification until the site tried to show a notification, even though the label said "You will be notified when this wiki is ready" (I didn't actually see a notification, only the permission prompt), that's not great.

Also, once I allowed them, I wasn't able to un-tick the button.

(I tested on patchdemo.wmflabs.org)

edg2s:
That's very weird. It does requestPermission when the button is toggled. Some browsers only allow permissions to be requested by a user-triggered event, but this is indeed that.

> Also, once I allowed them, I wasn't able to un-tick the button.

This is how the API works. You can't yet revoke permission from JS. The user has to do this in their browser.

edg2s:
It might be because I rebased out #185 which was probably required. Also in Chrome if you deny access enough times, it stops asking the user for permission, and will only just show "Notifications blocked" when it tries to send a notification. Maybe try again in a private window, or in a different browser.

edg2s:
I think this may be caused by https://blog.chromium.org/2020/01/introducing-quieter-permission-ui-for.html

edg2s:
The latest version puts the prompt on the index page. It's much less elegant, but what can you do...

edg2s:
> Also, once I allowed them, I wasn't able to un-tick the button.

I fixed this by storing a flag in `sessionStorage` to disable notifications without revoking the permission.

MatmaRex:
Thanks

PR #184, opened 2020-12-15
Minerva: Enable MinervaApplyKnownTemplateHacks

edg2s:
This applies some CSS that makes environments
more production-like.


edg2s:
CC @jdlrobson

jdlrobson:
Makes sense.

PR #191, opened 2020-12-23
Show progress bar as pending while wiki is being built

edg2s:
This ability was added upstream for this use case.

### DIFF
--- a/new.js
+++ b/new.js
@@ -6,6 +6,8 @@
 		document.getElementsByClassName( 'installProgressField' )[ 0 ]
 	);
 
+	pd.installProgressField.fieldWidget.pushPending();
+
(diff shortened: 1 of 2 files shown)


MatmaRex:
I don't really understand why, but alright.

edg2s:
With the non-pending progress bar, nothing on the page is moving most of the time. That does not give people confidence that things are happening. The alternative is to have the bar continuously and slowly moving towards to the next breakpoint, but that is fiddly to implement.

MatmaRex:
Hmm, fair enough. IMO there's the browser's loading indicator (usually some spinny thing near the tab), and also the install logs appearing, but this is fine.

PR #192, opened 2021-01-01
Add `if exists` guard to DROP DATABASE command

Ammarpad:
The command can sometimes be executed when the database is not yet created, currently this spits out errors because non-existing database cannot be deleted.


MatmaRex:
Thanks!

PR #195, opened 2021-01-08
Disable directory indexes on server

edg2s:


edg2s:
I've already applied this to the wmflabs install.

PR #203, opened 2021-01-21
Add an unintrusive anti-spam CAPTCHA

### LINKED ISSUE #202: Prevent spam
Disaster has struck. Who could have expected this?????

https://patchdemo.wmflabs.org/wikis/760d0aca2d743bb87426b5be2a3d15c7/w/index.php/Special:RecentChanges

(I only noticed because I checked the database sizes due to #199, and it turns out that wiki has 400 MB of spam on it already.)

For now, I disabled editing on that wiki. I'll set up something with ConfirmEdit.

MatmaRex:
Fixes #202

### DIFF
--- a/LocalSettings.txt
+++ b/LocalSettings.txt
@@ -83,3 +83,13 @@ $wgGENewcomerTasksTopicType = 'ores';
 $wgWelcomeSurveyExperimentalGroups['exp2_target_specialpage']['range'] = '0-9';
 $wgGEHomepageMentorsList = 'Project:GrowthExperiments_mentors';
 $wgGEHelpPanelHelpDeskTitle = 'Project:GrowthExperiments_help_desk';
+
+// Unintrusive anti-spam CAPTCHA
+if ( file_exists( 'extensions/ConfirmEdit/QuestyCaptcha/extension.json' ) ) {
+	wfLoadExtension( 'ConfirmEdit/QuestyCaptcha' );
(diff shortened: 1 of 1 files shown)


MatmaRex:
I added this to LocalSettings.php on the one wiki affected by spam so far, and it stopped (https://patchdemo.wmflabs.org/wikis/760d0aca2d743bb87426b5be2a3d15c7/w/index.php/Special:RecentChanges). Hopefully whoever is spamming us doesn't actually care.

jdforrester (approved):
Works well enough, but it's a bit irritating to do it on every edit.

MatmaRex:
Grant yourself sysop rights (or use the default "Patch Demo" account) and you won't see it.

MatmaRex:
In fact, you just need to be autoconfirmed, which I just found out because some of the spambots were already autoconfirmed, and were not getting the CAPTCHA. I set $wgAutoConfirmCount to a large number on that wiki, but it shouldn't be needed on any new wikis.

edg2s:
Makes the wikis slightly less production-like, but such is life.

It would be nice if we could try to patch this to old wikis, as they are more likely to get spammed.

edg2s:
Maybe going forward we should have `include('../../WikiFarmSettings.php');` in `LocalSettings.txt` so we can make batch changes to wiki configs more easily?

Issue #81, opened 2020-06-24
Repeated `Depends-On:` merged again


DemianX0:
Follow-up to #32. Non-breaking.
I've applied 2 patches, both `Depends-On:` the same patch. The latter was applied twice **without causing any issues**, but unnecessarily:
[![Screenshot from Gyazo](https://gyazo.com/fb85ad73ddad1bf9fe1dc391ee60454e/raw)](https://gyazo.com/fb85ad73ddad1bf9fe1dc391ee60454e)
http://patchdemo.wmflabs.org/wikis/87ff0eb77076404c5541ac861a7eb158/w/

edg2s:
From #204:
> Also, it looks like this kind of fixes #81 – duplicate patches are no longer shown in the table, because they can't be stored in the `patches` table (`patch` is a primary key).

I guess we are still applying the patch twice, but we aren’t showing it twice.

PR #204, opened 2021-01-22
Store wiki cache in structured DB instead of flat file

### LINKED ISSUE #113: wikicache.json can cache partially built wiki results
If someone visits index.php while a wiki is building, the metadata for a partially built wiki may get cached. We could not cache broken wikis, but that would mean we try to evaluate metadata for actually broken wikis on every view.

edg2s:
Fixes #113, #207

### DIFF
--- a/includes.php
+++ b/includes.php
@@ -66,6 +66,191 @@ echo '
 		</header>
 		<main>';
 
+$mysqli = new mysqli( 'localhost', 'patchdemo', 'patchdemo', 'patchdemo' );
+if ( $mysqli->connect_error ) {
+	die( $mysqli->connect_error );
+}
+
+function insert_wiki_data( string $wiki, string $creator, int $created, array $patches = [] ) {
+	global $mysqli;
+	$stmt = $mysqli->prepare( '
+		INSERT INTO wikis
+		(wiki, creator, created, patches)
+		VALUES(?, ?, FROM_UNIXTIME(?), ?)
+		ON DUPLICATE KEY UPDATE
+		patches = ?
+	' );
+	if ( !$stmt ) {
+		echo $mysqli->error;
+	}
+	$patches = json_encode( $patches );
+	$stmt->bind_param( 'ssiss', $wiki, $creator, $created, $patches, $patches );
+	$stmt->execute();
+	$stmt->close();
+}
+
+function wiki_add_announced_tasks( string $wiki, array $announcedTasks ) {
+	global $mysqli;
+	$stmt = $mysqli->prepare( 'UPDATE wikis SET announcedTasks = ? WHERE wiki = ?' );
+	$announcedTasks = json_encode( $announcedTasks );
+	$stmt->bind_param( 'ss', $accouncedTasks, $wiki );
+	$stmt->execute();
+	$stmt->close();
+}
+
+function get_wiki_data( string $wiki ) {
+	global $mysqli;
+
+	$stmt = $mysqli->prepare( '
+		SELECT wiki, creator, UNIX_TIMESTAMP( created ) created, patches, announcedTasks
+		FROM wikis WHERE wiki = ?
+	' );
+	if ( !$stmt ) {
+		echo $mysqli->error;
+	}
+	$stmt->bind_param( 's', $wiki );
+	$stmt->execute();
+	$res = $stmt->get_result();
+	$data = $res->fetch_assoc();
+	$stmt->close();
+
+	if ( !$data ) {
+		throw new Error( 'Wiki not found: ' . $wiki );
+	}
+
+	// Decode JSON
+	$data['patches'] = json_decode( $data['patches'] ) ?: [];
+	$data['announcedTasks'] = json_decode( $data['announcedTasks'] ) ?: [];
+
+	// Populate patch list
+	$patchList = [];
+	$linkedTasks = [];
+	if ( $data['patches'] ) {
+		foreach ( $data['patches'] as $patch ) {
+			[ $r, $p ] = explode( ',', $patch );
+			$patchData = get_patch_data( $r, $p );
+			$patchList[$patch] = $patchData;
+
+			get_linked_tasks( $patchData[ 'message' ], $linkedTasks );
+		}
+	}
+	$data['patchList'] = $patchList;
+
+	// Populate task list
+	$linkedTaskList = [];
+	foreach ( $linkedTasks as $task ) {
+		$linkedTaskList[$task] = get_task_data( $task );
+	}
+	$data['linkedTaskList'] = $linkedTaskList;
+
+	return $data;
+}
+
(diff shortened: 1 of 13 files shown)


MatmaRex:
I agree that this seems like a good thing to do, this has gotten too complicated to keep it in flat text files :(

Maybe there's a good reason for this in the code (which I haven't read yet), but why SQLite? We already have MySQL, you'd just need to `CREATE DATABASE` in setup.sh or something. `CREATE TABLE` and migrations should go in setup.sh either way.

edg2s:
Good point, I was thinking of the wiki dbs as separate but we already talk to them directly in a few places.

edg2s:
Fixed

sql/patchdemo.sql
```diff
@@ -0,0 +1,28 @@
+CREATE DATABASE IF NOT EXISTS patchdemo COLLATE "utf8_general_ci";
```
MatmaRex, line 1: Why `COLLATE "utf8_general_ci"`? I would just remove this part and use the default.

I might be out of date on this, but IIRC in MySQL, `utf8mb4` is the real Unicode, while `utf8` is BMP-only Unicode and it will explode in your face when you try to use non-BMP characters like emojis. Not sure how exactly a collation might explode, but still.

It seems that the default is `utf8mb4_0900_ai_ci` (https://dev.mysql.com/doc/refman/8.0/en/charset-applications.html), no idea what exactly it does, but it sounds reassuring.

edg2s: Sure - I'm just copying from old stuff, but if the defaults are sensible that's fine.

edg2s: Fixed

setup.sh
```diff
@@ -17,11 +17,11 @@ sudo -u www-data mkdir composer
 # Create folder for wikis
 sudo -u www-data mkdir wikis
 
-# create a database user that is allowed to create databases for each wiki
-sudo mysql -u root --password='' -e "
-CREATE USER 'patchdemo'@'localhost' IDENTIFIED BY 'patchdemo';
-GRANT ALL PRIVILEGES ON \`patchdemo\_%\`.* TO 'patchdemo'@'localhost';
-"
+# Create a database user that is allowed to create databases for each wiki,
+# and the central patchdemo database
```
MatmaRex, line 21: Duplicate comment (line 23)

includes.php
```diff
@@ -66,6 +66,191 @@
...
+	die( $mysqli->connect_error );
+}
+
+function insert_wiki_data( string $wiki, string $creator, int $created, array $patches = [] ) {
+	global $mysqli;
+	$stmt = $mysqli->prepare( '
+		INSERT INTO wikis
+		(wiki, creator, created, patches)
+		VALUES(?, ?, NOW(), ?)
+		ON DUPLICATE KEY UPDATE
```
MatmaRex, line 80: I don't like this but I think I'll propose my own change. (This looks like a good way to lose lots of data and in particular to reset the `created` date accidentally; but it looks like it needs a few changes elsewhere.)

MatmaRex: I see now that it only updates `patches`? (next line)

I missed it at first so that must mean it's very confusing. I still don't like it. :)

includes.php
```diff
@@ -66,6 +66,191 @@
...
+	$data['announcedTasks'] = json_decode( $data['announcedTasks'] ) ?: [];
+
+	// Populate patch list
+	$patchList = [];
+	$linkedTasks = [];
+	if ( $data['patches'] ) {
+		foreach ( $data['patches'] as $patch ) {
+			[ $r, $p ] = explode( ',', $patch );
+			$patchData = get_patch_data( $r, $p );
+			$patchList[$patch] = get_patch_data( $r, $p );
```
MatmaRex, line 132: Calling `get_patch_data()` twice will do SQL and Gerrit API queries twice?

includes.php
```diff
@@ -66,6 +66,191 @@
 		</header>
 		<main>';
 
+$mysqli = new mysqli( 'localhost', 'patchdemo', 'patchdemo', 'patchdemo' );
+if ( $mysqli->connect_error ) {
+	die( $mysqli->connect_error );
+}
+
+function insert_wiki_data( string $wiki, string $creator, int $created, array $patches = [] ) {
```
MatmaRex, line 74: `$created` is never used, which presumably causes the dates to be lost in the `migrateToSql.php` script.

sql/patchdemo.sql
```diff
@@ -0,0 +1,28 @@
+CREATE DATABASE IF NOT EXISTS patchdemo;
+
+USE patchdemo;
+
+CREATE TABLE IF NOT EXISTS wikis (
```
MatmaRex, line 5: We do queries with `ORDER BY created`, so there should be an index on it?

edg2s: If you want... but for 100 rows and index won't make much difference

MatmaRex:
So, how would I deploy this on patchdemo.wmflabs.org?

# Checkout this commit
# Run setup.sh to create database
# Run migrateToSql.php
# Carefully compare the outputs of index.php, and create some wikis

If something is broken, I just checkout master again, and it should be happy. Might need to drop the database manually, which should be fine too.

includes.php
edg2s: Fixed

includes.php
edg2s: Yeah - we should use some database abstraction framework probably. Writing queries gets out of hand fast.

includes.php
edg2s: Looks like a typo. Fixed.

setup.sh
edg2s: not quite - this says the user is allowed to create the central patchdemo database

the subsequent comment is actually creating it

I guess the comma is a bit ambiguous.

edg2s:
> If something is broken, I just checkout master again, and it should be happy. Might need to drop the database manually, which should be fine too.

Sounds about right. You should be able to revert at any point, although going forward we stop creating the metadata flat files (e.g. creator.txt)

sql/user.sql
```diff
@@ -0,0 +1,3 @@
+CREATE USER 'patchdemo'@'localhost' IDENTIFIED BY 'patchdemo';
+GRANT ALL PRIVILEGES ON `patchdemo%`.* TO 'patchdemo'@'localhost';
```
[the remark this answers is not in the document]
edg2s: Actually, I can add an IF NOT EXISTS to the CREATE USER

MatmaRex:
I deployed it, everything looks good!

MatmaRex:
Also, it looks like this kind of fixes #81 – duplicate patches are no longer shown in the table, because they can't be stored in the `patches` table (`patch` is a primary key).

PR #209, opened 2021-01-28
Add edit count page for spam monitoring

edg2s:
- Separate out a can_admin method
- Add edit count page for spam monitoring

### DIFF
--- a/editcounts.php
+++ b/editcounts.php
@@ -1,0 +1,63 @@
+<?php
+require_once "includes.php";
+
+if ( !can_admin() ) {
+	echo 'Access denied.';
+	if ( $useOAuth && !$user ) {
+		echo oauth_signin_prompt();
+	}
+	return;
+}
+
+$cache = load_wikicache();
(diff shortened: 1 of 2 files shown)


edg2s:
The page is pretty slow to generate when you have >100 wikis, which is why it is admin only for now.

Detected another spam-ridden wiki which I will delete: https://patchdemo.wmflabs.org/wikis/04e7427feb3734e10d56f9ea3af7e4c2/w/index.php/Special:RecentChanges?hidebots=1&limit=50&days=7&enhanced=1&urlversion=2

image

MatmaRex:
By "pretty slow" you mean it takes a minute and a half, which is a little crazy. I wonder why, is the overhead of starting up MediaWiki to run the maintenance script so high? Maybe we should just do the MySQL queries ourselves (the script basically does `select * from site_stats`). I'll experiment with it, but let's do this for now.

MatmaRex:
#215

PR #216, opened 2021-01-31
Use protocol-relative $wgServer for the wikis

MatmaRex:
Just to be fancy.

### DIFF
--- a/new.php
+++ b/new.php
@@ -14,7 +14,7 @@ $patches = trim( $_POST['patches'] );
 $announce = !empty( $_POST['announce'] );
 
 $namePath = md5( $branch . $patches . time() );
-$server = detectProtocol() . '://' . $_SERVER['HTTP_HOST'];
+$server = '//' . $_SERVER['HTTP_HOST'];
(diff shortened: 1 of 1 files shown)


edg2s (approved):

new.php
```diff
@@ -298,9 +299,9 @@ function set_progress( int $pc, string $label ) {
 				[
 					'type' => 'comment',
 					'value' =>
-						"Test wiki created on [[ $server$serverPath | Patch Demo ]]" . ( $user ? ' by ' . $user->username : '' ) . " using patch(es) linked to this task:\n" .
+						"Test wiki created on [[ $protocol$server$serverPath | Patch Demo ]]" . ( $user ? ' by ' . $user->username : '' ) . " using patch(es) linked to this task:\n" .
 						"\n" .
-						"$server$serverPath/wikis/$namePath/w/"
+						"$protocol$server$serverPath/wikis/$namePath/w/"
```
edg2s, line 304: Actually, I think this is missing the ':'

edg2s:
I'm not sure this is worth it, it just means we have two versions of the URL around which can lead to bugs. The full URL works fine in all situations.

MatmaRex:
Eh, fair enough.

PR #234, opened 2021-02-05
Store branch name when creating wikis

edg2s:


MatmaRex:
I wouldn't even have thought about querying the `text` table with SQL, I got so used to the WMF setup where that table is on a completely different server (and also the entries are compressed). I was expecting this to read the branches from Git. But this works perfectly :D

Issue #196, opened 2021-01-19
Warn users when they try to create a duplicate wiki 


iamjessklein:
When I'm creating a patch on testdemo I want to be warned if the wiki has already been made and given a link to the duplicate. I should still be able to make another wiki, I just want that info .... and also not to have to wait forever for the wiki to be built. 

I think this could be a simple little dialog.

edg2s:
The slight problem is that we have to do API requests to know if the patches are duplicates (we need to know if there is a newer patchset for a given ID). We could either do the API request in the client, or put the warning on the create page as an interstitial.

We should also warn if the user is trying to create a wiki with no patches, which is much easier to detect.

iamjessklein:
+1 @edg2s

edg2s:
With the above patch:

> ![image](https://user-images.githubusercontent.com/180672/112489218-23ba4780-8d76-11eb-9be8-ebe7ae8714ac.png)

Clicking on the wiki link takes you to the table with the row highlighted:
> ![image](https://user-images.githubusercontent.com/180672/112489306-3af93500-8d76-11eb-9c3f-2ed432c9c609.png)

PR #235, opened 2021-02-07
Warn when creating a wiki with no patches

edg2s:
Part of #196


MatmaRex:
I'm not sure if this is helpful or annoying… I create wikis with no patches all the time.

edg2s:
It’s an edge case and only requires one more click.

MatmaRex:
Ah, whatever.

PR #239, opened 2021-02-18
Add Popups extension

edg2s:
We already have TextExtracts and PageImages which are semi-optional dependencies.


edg2s:
Verified locally:
image

PR #240, opened 2021-02-18
Allow use of the Mobilefrontend content provider proxy

jdlrobson:
Untested, but I'd like to add a proxy checkbox for testing languages as part of the desktop improvements project. It's really hard to test these without and I've been having to workaround this by using https://gerrit.wikimedia.org/r/c/mediawiki/extensions/MobileFrontend/+/641808

### DIFF
--- a/LocalSettings-proxy.txt
+++ b/LocalSettings-proxy.txt
@@ -1,0 +1,5 @@
+// Allow proxying content from production
+
+$wgMFContentProviderClass = "MobileFrontend\\ContentProviders\\MwApiContentProvider";
+$wgMFContentProviderTryLocalContentFirst = true;
+$wgMFAlwaysUseContentProvider = true;
--- a/index.php
+++ b/index.php
@@ -147,6 +147,19 @@ if ( $useOAuth && !$user ) {
 						]
 					),
 					new OOUI\FieldLayout(
+						new OOUI\CheckboxInputWidget( [
+							'name' => 'proxy',
+							'value' => 1,
+							'selected' => false
+						] ),
+						[
+							'label' => 'Proxy articles from en.wikipedia.org',
+							'help' => 'Any articles not local to the wiki will be pulled from English Wikipedia.',
+							'helpInline' => true,
+							'align' => 'left',
+						]
+					),
+					new OOUI\FieldLayout(
 						new OOUI\ButtonInputWidget( [
 							'classes' => [ 'form-submit' ],
 							'label' => 'Create demo',
--- a/new.php
+++ b/new.php
@@ -248,10 +248,18 @@ foreach ( $linkedTasks as $task ) {
 // Choose repositories to enable
 $repos = get_repo_data();
 
+$useProxy = (int)$_POST['proxy'];
+
 if ( $_POST['preset'] === 'custom' ) {
 	$allowedRepos = $_POST['repos'];
 } else {
 	$allowedRepos = get_repo_presets()[ $_POST['preset'] ];
+}
+
+// When proxying, always enable MobileFrontend
+if ( $useProxy ) {
+	// Doesn't matter if this appears twice
+	$allowedRepos[] = 'mediawiki/extensions/MobileFrontend';
 }
 
 foreach ( array_keys( $repos ) as $repo ) {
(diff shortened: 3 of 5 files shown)


index.php
```diff
@@ -146,6 +146,19 @@
...
 					),
+					new OOUI\FieldLayout(
+						new OOUI\CheckboxInputWidget( [
+							'name' => 'proxy',
+							'value' => 1,
+							'selected' => false
+						] ),
+						[
+							'label' => 'Proxy articles from en.wikipedia.org',
+							'help' => 'Any content not locally to the wiki will be pulled from production wiki	',
```
edg2s, line 157: content->articles
locally->local

or:
"Any articles which don't exist on the wiki will be fetched from English Wikipedia"

edg2s:
This seems generally useful enough to me. We may need to consider an "advanced options" section in the near future.

index.php
```diff
@@ -146,6 +146,19 @@
 							'classes' => [ 'form-repos-field' ],
 						]
 					),
+					new OOUI\FieldLayout(
+						new OOUI\CheckboxInputWidget( [
+							'name' => 'proxy',
+							'value' => 1,
+							'selected' => false
```
edg2s, line 153: Should probably be disabled by default.

jdlrobson: Disabled or checked (selected)?

edg2s: Sorry, unchecked (selected=false)

jdlrobson: Sorry.. still a little confused. what do you want me to change here? You want me to make it selected = true ?

index.php
jdlrobson: (done)

edg2s:
If the purpose is to test content rendering it would probably be useful to also import en:MediaWiki:Common.css?

One could either fetch the latest version in PHP, or just add a getCSS call to Common.js (provided there are no CORS issues).

createwiki.sh
```diff
@@ -48,6 +48,11 @@ php $PATCHDEMO/wikis/$NAME/w/maintenance/install.php \
 # apply our default settings
 cat $PATCHDEMO/LocalSettings.txt >> $PATCHDEMO/wikis/$NAME/w/LocalSettings.php
 
+# Add the proxy if selected
+if [ "${USE_PROXY}" = "1" ] then
+	cat $PATCHDEMO/wikis/$NAME/w/LocalSettings.php $PATCHDEMO/LocalSettings-proxy.txt
```
MatmaRex, line 53: Missing `>>`?

new.php
```diff
@@ -236,6 +236,8 @@ function set_progress( int $pc, string $label ) {
 // Choose repositories to enable
 $repos = get_repo_data();
 
+$useProxy = $_POST['proxy'];
```
MatmaRex, line 239: Add `(int)` or some other validaiton thing

createwiki.sh
edg2s: and they're the wrong way around
`cat txt >> php`

jdlrobson: (done)

new.php
jdlrobson: (done)

createwiki.sh
```diff
@@ -47,3 +47,8 @@ php $PATCHDEMO/wikis/$NAME/w/maintenance/install.php \
 
 # apply our default settings
 cat $PATCHDEMO/LocalSettings.txt >> $PATCHDEMO/wikis/$NAME/w/LocalSettings.php
+
+# Add the proxy if selected
+if [ "${USE_PROXY}" = "1" ] then
+	cat $PATCHDEMO/wikis/$NAME/w/LocalSettings.php $PATCHDEMO/LocalSettings-proxy.txt >> $PATCHDEMO/wikis/$NAME/w/LocalSettings.php
```
edg2s, line 53: >> means append, so you don't need to repeat the php file. I'll fix in the rebase

LocalSettings-proxy.txt
```diff
@@ -0,0 +1,5 @@
+// Allow proxying content from production
+
+$wgMFContentProviderClass = "MobileFrontend\\ContentProviders\\MwApiContentProvider";
+$wgMFContentProviderTryLocalContentFirst = false;
```
edg2s, line 4: This option should be dropped then, given what we say in the form below. Otherwise we would always get remote content.

jdlrobson: Switched to true

edg2s:
This should also enforce that MobileFrontend is selected in the extensions list.

edg2s (approved):

MatmaRex:
Wild.

PR #241, opened 2021-02-22
Add StopForumSpam extension to patchdemo

sbassett29:
https://phabricator.wikimedia.org/T271991

### DIFF
--- a/repositories-preset-wikimedia.txt
+++ b/repositories-preset-wikimedia.txt
@@ -85,6 +85,7 @@ mediawiki/extensions/SecureLinkFixer
 mediawiki/extensions/SecurePoll
 mediawiki/extensions/SiteMatrix
 mediawiki/extensions/SpamBlacklist
+mediawiki/extensions/StopForumSpam
 mediawiki/extensions/SyntaxHighlight_GeSHi
 mediawiki/extensions/TemplateData
 mediawiki/extensions/TemplateSandbox
--- a/repositories.txt
+++ b/repositories.txt
@@ -36,6 +36,7 @@ mediawiki/extensions/ReplaceText w/extensions/ReplaceText
 mediawiki/extensions/Scribunto w/extensions/Scribunto
 mediawiki/extensions/SecurePoll w/extensions/SecurePoll
 mediawiki/extensions/SpamBlacklist w/extensions/SpamBlacklist
+mediawiki/extensions/StopForumSpam w/extensions/StopForumSpam
 mediawiki/extensions/SyntaxHighlight_GeSHi w/extensions/SyntaxHighlight_GeSHi
 mediawiki/extensions/TemplateData w/extensions/TemplateData
 mediawiki/extensions/TemplateStyles w/extensions/TemplateStyles


repositories-preset-wikimedia.txt
```diff
@@ -85,6 +85,7 @@ mediawiki/extensions/SecureLinkFixer
 mediawiki/extensions/SecurePoll
 mediawiki/extensions/SiteMatrix
 mediawiki/extensions/SpamBlacklist
+mediawiki/extensions/StopForumSpam
```
edg2s, line 88: I don't know much about this extension, is it installed on any WMF wikis?

MatmaRex:
Looks like this extension mostly disallows edits from specific IP addresses. It might not really work, because the Patchdemo wikis see all edits as coming from private Wikimedia Cloud IPs, because I never figured out how to set that up properly. So you might need to figure out that for me ;)

repositories-preset-wikimedia.txt
sbassett29: It's allegedly used by a handful of wikis right now: https://wikiapiary.com/wiki/Extension:StopForumSpam

It's also on Wikimedia beta right now, in report-only mode, with the eventual goal of deploying to production (or at least to a few projects), hopefully sooner than later: https://phabricator.wikimedia.org/T273220

sbassett29:
Yeah, it's certainly not a necessity that SFS be available in patchdemo.  I just wanted to add it to have an additional, convenient way to test parts of it.  I get that it's not a UI-heavy extension, which is maybe more the ideal candidate for patchdemo inclusion.  If you want to decline this for now, that's fine.

MatmaRex:
No, that's just fine.

However, I tested this, and it blows up: "Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /srv/patchdemo-wikis/2a0ba4c07903db0f92bbbfd4634a5e84/w/vendor/wikimedia/ip-set/src/IPSet.php on line 178"
https://patchdemo.wmflabs.org/wikis/2a0ba4c07903db0f92bbbfd4634a5e84/w/index.php?title=Test&action=edit&redlink=1

Must be a lot of IPs… I guess we need to bump the memory limits in PHP configuration somewhere.

sbassett29:
Hmm, the default SFS deny list (https://www.stopforumspam.com/downloads/listed_ip_90_ipv46_all.gz) is about 7.6 Mb uncompressed.  I wouldn't assume that amount of data would cause oom errors very easily, but maybe there's some inefficiency with how IPSet is trying to match on the wancache data that shows up under wmfcloud instances.

MatmaRex:
Maybe it's just straw that broke the camel's back. Either way I'm increasing the limit in #259, and that seems to make it work.

edg2s:
When does this run? I'm a bit concerned about something that caused us to increase the memory limit being enabled by default.

sbassett29:
@edg2s For the primary cache update, there is a [maint script](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/StopForumSpam/+/refs/heads/master/maintenance/updateDenyList.php) that can be run manually or cronned, but that's not really relevant for patchdemo.  The main trigger happens in the [onGetUserPermissionsErrorsExpensive hook](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/StopForumSpam/+/refs/heads/master/includes/Hooks.php#115), specifically [where the deny list update is added as a deferred update](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/StopForumSpam/+/refs/heads/master/includes/Hooks.php#139).

MatmaRex:
The extension is causing trouble and I removed it from all existing wikis and disabled it by default, see #284 and https://phabricator.wikimedia.org/T279449. You can still enable it for testing if you want.

PR #249, opened 2021-02-25
Enable inbox extension for testing emails

edg2s:


MatmaRex:
I was actually just about to click the button here

edg2s:
:) Thought, I'd bundle this with the extension that will use it (although the extension doesn't exist in Gerrit yet), so ... -1

MatmaRex:
Patchdemo actually sends you emails currently (I just tested it, was a bit surprised that it works), so this would have still had an effect. But we can do them together as well.

edg2s:
Interesting, must be a cloud vps thing. Moved to #251.

MatmaRex:
The repo has been imported: https://www.mediawiki.org/w/index.php?title=Gerrit/New_repositories/Requests/Entries&diff=prev&oldid=4427762&diffmode=source

PR #256, opened 2021-02-28
Use short URLs

### LINKED ISSUE #248: Configure Short URLs
https://www.mediawiki.org/wiki/Manual:Short_URL

This would make Patch Demo wikis a bit more production-like.

edg2s:
Fixes #248

### DIFF
--- a/LocalSettings.txt
+++ b/LocalSettings.txt
@@ -11,6 +11,9 @@ $wgDebugComments = true;
 // Don't share cookies across wikis (#78)
 $wgCookiePath = $wgScriptPath;
 
+// Short URLs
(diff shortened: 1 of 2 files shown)


MatmaRex:
The config looks like the recommended one from <https://www.mediawiki.org/wiki/Manual:Short_URL/Apache>, I like that.

MatmaRex:
I just realized that this breaks logging in. `$wgCookiePath` points to `/w`, so the cookies are not sent when visiting `/wiki/…`.

I guess that's a good opportunity to improve the "iki" thing you did, which I wasn't going to complain about but it was a bit super ugly.

PR #258, opened 2021-03-02
Set up for custom SQL being run for new wikis

kemayo:
This is a quick sketch of what'd probably work. I don't have patchdemo set up locally to test it properly.

### DIFF
--- a/new/postinstall.sh
+++ b/new/postinstall.sh
@@ -23,6 +23,12 @@ php $PATCHDEMO/wikis/$NAME/w/maintenance/resetUserEmail.php --no-reset-password 
 php $PATCHDEMO/wikis/$NAME/w/maintenance/resetUserEmail.php --no-reset-password "Bob" Bob@localhost
 php $PATCHDEMO/wikis/$NAME/w/maintenance/resetUserEmail.php --no-reset-password "Mallory" Mallory@localhost
 
+# run arbitrary SQL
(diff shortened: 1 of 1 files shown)


kemayo:
In theory the sql-perwiki directory would come to contain things like the interwiki table config, per @jdlrobson.

new/postinstall.sh
```diff
@@ -17,6 +17,12 @@ php $PATCHDEMO/wikis/$NAME/w/maintenance/createAndPromote.php Mallory patchdemo1
 # This command may fail as --disable-autoblock was only added in 1.36, so suppress errors
 echo "Mallory" | php $PATCHDEMO/wikis/$NAME/w/maintenance/blockUsers.php --reason "Blocking account for testing" --disable-autoblock || echo "Can't block Mallory"
 
+# run arbitrary SQL
+for sql in $(find $PATCHDEMO/sql-perwiki -name "*.sql" -not -type d -printf '%P\n')
+do
+	mysql -u patchdemo -ppatchdemo patchdemo < $PATCHDEMO/sql-perwiki/$sql
```
edg2s, line 23: The database name for the wiki is patchdemo_$NAME

edg2s (changes requested):

new/postinstall.sh
kemayo: Fixed

edg2s (approved):
Tested locally, appears to work.

PR #259, opened 2021-03-02
Increase PHP memory_limit

MatmaRex:
The default is not enough for the StopForumSpam extension. (#241)

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -36,9 +36,15 @@ echo "#!/bin/bash
 sudo mysqlcheck --optimize --all-databases -u root --password=''" > /etc/cron.monthly/patchdemo-optimize
 chmod u+x /etc/cron.monthly/patchdemo-optimize
 
-# set session expiration to a month (default is 24 minutes???), cookie expiration too
-echo "session.gc_maxlifetime = 2592000
-session.cookie_lifetime = 2592000" > /etc/php/7.3/apache2/conf.d/patchdemo.ini
+# PHP settings
+echo "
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -36,9 +36,15 @@ echo "#!/bin/bash
...
-session.cookie_lifetime = 2592000" > /etc/php/7.3/apache2/conf.d/patchdemo.ini
+# PHP settings
+echo "
+; set session expiration to a month (default is 24 minutes???), cookie expiration too
+session.gc_maxlifetime = 2592000
+session.cookie_lifetime = 2592000
+
+; double the default memory limit
+memory_limit = 256M
+" > /etc/php/7.3/apache2/conf.d/patchdemo.ini
```
edg2s, line 47: This seems flaky. I know `php -ini` gets you the path to the ini files used in the CLI, which would be some improvement (it would only break if you had multiple versions of PHP installed)

edg2s (changes requested):

setup.sh
MatmaRex: I don't think it's flaky, PHP configuration has worked this way since forever. It's documented too: <https://wiki.debian.org/PHP#Configuration_layout>. The only thing we'll need to update is the version number, some day, but we can think about that when Cloud instances will be moving to Debian 11. (And it should fail loudly in that case, since the directory won't exist.)

edg2s: I mean the version number, but sure.

PR #260, opened 2021-03-02
Run the populateInterwiki maintenance script

edg2s:
By default this imports interwikis from en.wiki


edg2s:
It's takes a second or two (depending on network conditions)

PR #267, opened 2021-03-03
Improve patch number validation

edg2s:


MatmaRex:
Previously it also allowed SHA1 commit hashes, which was intended, but that stopped working in 202df39fb744fe2e6b3a2a82997c6b251a5d8ab4.

PR #268, opened 2021-03-03
Create a PatchSelectWidget

edg2s:
Based on TagMultiselectWidget, shows the patch number and metadata when a valid patch is entered.

### DIFF
--- a/PatchSelectWidget.php
+++ b/PatchSelectWidget.php
@@ -1,0 +1,6 @@
+<?php
+
+use OOUI\MultilineTextInputWidget;
+
+class PatchSelectWidget extends MultilineTextInputWidget {
+}
--- a/api.php
+++ b/api.php
@@ -1,0 +1,18 @@
+<?php
+
+require_once "includes.php";
+
+header( 'Content-Type: application/json' );
+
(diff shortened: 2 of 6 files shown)


PatchSelectWidget.js
```diff
@@ -0,0 +1,161 @@
+/* global OO, $ */
+
+window.PatchSelectWidget = function PatchSelectWidget( config ) {
+	var widget = this;
+
+	// Parent constructor
+	window.PatchSelectWidget.super.call( this, $.extend( {
+		allowArbitrary: true,
+		inputPosition: 'outline',
```
edg2s, line 9: I prefer the look of inline to outline, but also I want each task to be on a separate line so they don't jump when the names arrive from the api.

edg2s:
This should make #196 a bit easier.

PatchSelectWidget.js
edg2s: I've pushed an inline version on top of this, maybe the jumping is ok?

edg2s:
Inline version:
image

Outline version:
image

MatmaRex:
Looks way too professional. ;)

PR #271, opened 2021-03-06
Add typehints to all PHP functions

edg2s:


includes.php
```diff
@@ -344,7 +344,7 @@ function delete_wiki( $wiki ) {
 
 $requestCache = [];
 
-function gerrit_query( $url, $echo = false ) {
+function gerrit_query( string $url, $echo = false ) : string {
```
MatmaRex, line 347: Returns an array, not string

PR #275, opened 2021-03-11
Use a 10 character hash for wiki names

edg2s:
This makes URLs and other things a bit cleaner by not
using all 32 characters.

16^10 is still a trillion, so unlikely to cause a collision.

Truncate the display of older hashes so they look consistent.


edg2s:
Collisions only become likely when you have ~sqrt(N) items, so in this case you'd need a million wikis.

MatmaRex:
Neat.

PR #277, opened 2021-03-12
Add empty line after checkboxes and before warning

### LINKED ISSUE #276: Incorrect alignment of the warning
When you have created a wiki without specifying the patch or a wiki is created with the merged patch(es), there is a warning.

There isn't an empty line before warning and after checkbox for "Previously generated wikis".

See:
image

I have a fix for this:
image

kizule:
Fixes #276.

### DIFF
--- a/css/common.css
+++ b/css/common.css
@@ -98,6 +98,10 @@ summary > .oo-ui-labelElement-label {
 
 /* Wiki list */
 
+.oo-ui-messageWidget.oo-ui-messageWidget-block.oo-ui-flaggedElement-warning {
+	margin-top: 16px;
+}
+
 .showClosed {
 	float: right;
 	margin-top: -6px;
--- a/css/common.css
+++ b/css/common.css
@@ -98,13 +98,8 @@ summary > .oo-ui-labelElement-label {
 
 /* Wiki list */
 
-.oo-ui-messageWidget.oo-ui-messageWidget-block.oo-ui-flaggedElement-warning {
-	margin-top: 16px;
-}
-
 .showClosed {
 	float: right;
-	margin-top: -6px;
 }
 
 .status-ABANDONED:before {


css/common.css
```diff
@@ -98,6 +98,10 @@ summary > .oo-ui-labelElement-label {
 
 /* Wiki list */
 
+.oo-ui-messageWidget.oo-ui-messageWidget-block.oo-ui-flaggedElement-warning {
+	margin-top: 16px;
```
edg2s, line 102: use 1em for the margin

edg2s, line 102: Use 1em for the margin

css/common.css
```diff
@@ -98,6 +98,10 @@ summary > .oo-ui-labelElement-label {
 
 /* Wiki list */
 
+.oo-ui-messageWidget.oo-ui-messageWidget-block.oo-ui-flaggedElement-warning {
```
edg2s, line 101: Add a new class to the widget using `'classes' => [ 'closedWikiWarning' ]`, then just use that class as the selector here: `.closedWikiWarning {`

edg2s (changes requested):

css/common.css
kizule: I found that .showClosed had -6px for margin-top.
In index.php I found that this element is for the warning box. I've removed it from common.css

css/common.css
kizule: See the previous comment.

css/common.css
```diff
@@ -100,7 +100,6 @@ summary > .oo-ui-labelElement-label {
 
 .showClosed {
 	float: right;
-	margin-top: -6px;
```
edg2s, line 103: This negative margin is for the alignment of the button within the warning box, not the box itself.

edg2s (changes requested):

MatmaRex:
It doesn't work, I think `.showClosed` is the wrong element.

kizule:
I don't know how to do this, so someone else should take a look in this.

edg2s:
See #283

PR #280, opened 2021-03-14
Suppress warnings from file_get_contents

edg2s:


MatmaRex:
Does this do anything? Won't you get a warning on the next line when you try to `json_decode(false)`? (Or an exception somewhere later?)

MatmaRex:
Oh I see, it fixes the api.php thing from #268.

PR #287, opened 2021-04-21
Fix display of marker on <summary> tag

edg2s:
OOUI sets display to block, when it should be list-item.


MatmaRex:
And here I thought that nothing in HTML or CSS could surprise me any more.

edg2s:
It's 2003, everything is a list!

PR #295, opened 2021-05-05
Start reducing the code in the global scope

DannyS712:
Create a helper class, PatchDemoWikiDatabase, to handle interactions with the `wiki` table, and start moving logic there.


MatmaRex:
Sorry, I don't really feel like this is an improvement… I know that this is a good practice and all, but this is a simple project, and I prefer to keep the code simple.

DannyS712:
> Sorry, I don't really feel like this is an improvement… I know that this is a good practice and all, but this is a simple project, and I prefer to keep the code simple.

okay

PR #297, opened 2021-05-05
Separate config files by extension/skin/service and load conditionally

edg2s:


localsettings/core.txt
```diff
@@ -0,0 +1,46 @@
...
+		'width' => 190,
+		'height' => 50,
+	]
+];
+$wgFavicon = "$wgResourceBasePath/favicon.ico";
+
+// Email settings
+$wgAllowHTMLEmail = true;
+
+// Interwiki
```
DannyS712, line 45: Should this be in an extension-specific file too?

localsettings/extensions-DiscussionTools.txt
```diff
@@ -0,0 +1,2 @@
+$wgLocaltimezone = 'UTC';
```
DannyS712, line 1: since these configuration settings are for settings in core, I would suggest putting them in the overall settings that are loaded each time, to have a more stable interface across different patchdemo wikis - the other configuration files are for extension settings

localsettings/services-parsoid.txt
```diff
@@ -0,0 +1,18 @@
+// Parsoid/PHP required configuration
+$wgEnableRestAPI = true;
```
DannyS712, line 2: was going to suggest putting $wgEnableRestAPI in the overall core settings per above, but that configuration has been removed from core, so might not be needed at all

MatmaRex: Most of the Parsoid section is no longer needed for latest MediaWiki, but I wanted to keep it, because Patchdemo allows creating wikis using older MediaWiki versions too. See #82.

edg2s:
Separating config by role is pretty standard, I don't think it has to necessarily be a slippery slope to something more.

localsettings/core.txt
edg2s: Yes.

localsettings/extensions-DiscussionTools.txt
edg2s: Done

PR #301, opened 2021-06-08
Enable watchlist expiry feature

edg2s:
This is enabled on all production wikis

### DIFF
--- a/localsettings/core.txt
+++ b/localsettings/core.txt
@@ -49,4 +49,4 @@ $wgFragmentMode = [ 'html5', 'legacy' ];
 $wgLocaltimezone = 'UTC';
 
 // Enable watchlist expiry feature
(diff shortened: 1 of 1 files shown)


edg2s:
See also https://phabricator.wikimedia.org/T284563

localsettings/core.txt
```diff
@@ -47,3 +47,6 @@ $wgFragmentMode = [ 'html5', 'legacy' ];
 
 // Choose a sensible timezone
 $wgLocaltimezone = 'UTC';
+
+// Enable watchlist expiry feature
+$wgWatchlistExpiry = false;
```
Ammarpad, line 52: The commit summary says 'enable', I guess this should be $wgWatchlistExpiry = true;?

edg2s: oops

PR #305, opened 2021-07-07
GrowthExperiments: Link suggestion support

kostajh:
Use static task suggester for suggested edits, and include content for
two article pages and link suggestions for one of those articles.

Bug: T280664

### DIFF
--- a/localsettings/extensions-GrowthExperiments.txt
+++ b/localsettings/extensions-GrowthExperiments.txt
@@ -4,3 +4,8 @@ $wgGENewcomerTasksTopicType = 'ores';
 $wgWelcomeSurveyExperimentalGroups['exp2_target_specialpage']['range'] = '0-9';
 $wgGEHomepageMentorsList = 'Project:GrowthExperiments_mentors';
 $wgGEHelpPanelHelpDeskTitle = 'Project:GrowthExperiments_help_desk';
+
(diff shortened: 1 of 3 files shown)


MatmaRex:
I'm not really sure how to test this (I get the prompt to add links, but not actual suggestions), so I'll just assume that you know and it works for you.

kostajh:
> I'm not really sure how to test this (I get the prompt to add links, but not actual suggestions), so I'll just assume that you know and it works for you.

Thanks!

Do you mind sending a screenshot and/or looking at browser console to see if there are errors? I tested with the patchdemo instance you created. I made a new account, went to Special:Homepage, clicked on "Douglas Adams" and saw the onboarding dialog and then the AddLink plugin loaded. If you were testing with the default Admin user, there might have been some settings not set for that user.

MatmaRex:
Oh, it worked as expected after I created a new account. I wasn't expecting that to make a difference, but I guess it makes sense since the extension is supposed to do stuff for new users only. It might make testing annoying though, since you have to click through a bunch of dialogs.

I was previously testing on https://patchdemo.wmflabs.org/wikis/59eb222e6e logged in as Bob. I enabled the newcomer homepage in preferences (it wasn't on by default), then followed the same steps as you, but after clicking the link I just got a standard visual editor. (No errors.) You'll probably be able to reproduce it if you log in with that account.

PR #308, opened 2021-08-05
Use the new MobileFrontendContentProvider

jdlrobson:
Add the extension to the list but only enable it if proxy
is set or its added in the dropdown.

Drop configuration variables as these are all the defaults
in the new extension.

Should not be merged until https://gerrit.wikimedia.org/r/c/mediawiki/extensions/MobileFrontend/+/709544
has been merged.

### DIFF
--- a/localsettings/feature-proxy.txt
+++ b/localsettings/feature-proxy.txt
@@ -1,8 +1,7 @@
 // Allow proxying content from production
 
-$wgMFContentProviderClass = "MobileFrontend\\ContentProviders\\MwApiContentProvider";
+$wgMFContentProviderClass = "MobileFrontendContentProviders\\MwApiContentProvider";
 $wgMFContentProviderTryLocalContentFirst = true;
-$wgMFAlwaysUseContentProvider = true;
 
 // Enable proxying for Page previews if enabled.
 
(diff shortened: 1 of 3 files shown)


MatmaRex:
Looks OK, please ping me when it should be deployed.

jdlrobson:
Okay, consider yourself pinged! This is ready to be merged.

MatmaRex:
I applied this on https://patchdemo.wmflabs.org/ and it doesn't seem to actually work. The content is not proxies on this wiki I just created: https://patchdemo.wmflabs.org/wikis/4e403397fd/wiki/Dog

MatmaRex:
Oh, I see why:
```
Cloning into 'mediawiki/extensions/MobileFrontendContentProvider'...
fatal: remote error: mediawiki/extensions/MobileFrontendContentProvider unavailable
```

MatmaRex:
mediawiki/extensions/MobileFrontendContentProvider is not on Gerrit, as far as I can tell. I think you need to get that set up first, Patchdemo can only fetch stuff from Gerrit.

jdlrobson:
Okay I guess we are waiting on https://www.mediawiki.org/wiki/Gerrit/New_repositories/Requests then.

MatmaRex:
The repo was created and this seems to work now. Thanks!

jdlrobson:
Thanks for the merge!

PR #310, opened 2021-09-15
Fix test page for Disambiguator extension

MusikAnimal:
The actual dab page was missing the `__DISAMBIG__` magic word


MatmaRex:
I wanted to test this and apparently something else broke preventing new wikis from being created, see #311.

MusikAnimal:
> I wanted to test this and apparently something else broke preventing new wikis from being created, see #311.

Yeah I noticed that and was about to report it here, but wasn't sure if it was an issue with Patch Demo or the AbuseFilter itself. I just deselected AbuseFilter from the extensions list, but looks like you've already fixed the issue.

Thanks for merging, and for all the work you and others have put into Patch Demo!

PR #313, opened 2021-09-29
Set empty actions cell after loop

### LINKED ISSUE #312: On the main page, wikis newer than the last I created have missing cells
If I’ve created at least one wiki, there’s an extra Actions column containing the _Delete_ link. In rows below it that belong to other people’s wikis, this cell is empty. However, in rows above it, the cell doesn’t exist at all.

Screenshot_2021-09-28 Patch demo

This is because `index.php` loops over the data only once. Instead, it should loop once to read from the database and preprocess the data (including setting `$anyCanDelete`), and a second time actually create the HTML, reading `$anyCanDelete` only this second time. (The second run could be placed after printing the table header, so that instead of assembling a string, it could write directly on stdout.)

edg2s:
Fixes #312

### DIFF
--- a/index.php
+++ b/index.php
@@ -298,7 +298,7 @@ while ( $data = $results->fetch_assoc() ) {
 		( $canAdmin ? '<td data-label="Time to create">' . ( $wikiData['timeToCreate'] ? $wikiData['timeToCreate'] . 's' : '' ) . '</td>' : '' ) .
 		( $canDelete ?
 			'<td data-label="Actions"><a href="delete.php?wiki=' . $wiki . '">Delete</a></td>' :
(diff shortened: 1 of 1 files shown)


MatmaRex (changes requested):
This lets anyone mess up the markup (or worse) by adding "__EMPTY_ACTIONS__" to the Phab task title or some other places.

I'm fine with doing this the lazy way, but the placeholder needs to be something that can't appear in user input (e.g. wrap it in `<!-- -->`, because any angle brackets in user input would have been escaped, so they're safe).

edg2s:
Done

MatmaRex (approved):

tacsipacsi:
Thanks for the fix!

PR #319, opened 2021-10-09
Add MassMessage to allowed extensions

DannyS712:
WMF deployed extension that should be available to test.
Already included in repository-lists/wikimedia.txt but not actually available in the list of supported extensions.


MatmaRex:
Thanks, looks like it works out-of-the-box.

PR #340, opened 2021-11-08
Logo: Set font to Montserrat

edg2s:
Also split logo in CSS so text can be coloured
on hover, and hidden on mobile.


MatmaRex:
Whoa…

PR #341, opened 2021-11-09
build: Update devDependencies

edg2s:


MatmaRex:
The change in CSS is surprising, I guess stylelint forbids `px` now?

PR #342, opened 2021-11-09
README: Add header, fix casing of "Patch demo" and bold

edg2s:


MatmaRex:
I updated the repo description and https://toolhub.wikimedia.org/tools/patchdemo to match. I haven't really paid attention to the casing before, so there might be other places where it's inconsistent.

MatmaRex:
Actually, the wikis themselves have "Patch Demo" in uppercase on the main page and in the title, and the default user is also "Patch Demo". I don't care so I'm not changing it.

edg2s:
> I updated the repo description and https://toolhub.wikimedia.org/tools/patchdemo to match. I haven't really paid attention to the casing before, so there might be other places where it's inconsistent.

I think you can provide a URL for the toolhub config, which we can host here.

PR #343, opened 2021-11-09
Store notification toggle state in localStorage

edg2s:
There's no reason for this preference to not persist
across multiple sessions.


MatmaRex (changes requested):
Need to update new.js  as well?

edg2s:
> Need to update new.js as well?

Oops, fixed.

MatmaRex (approved):

PR #347, opened 2021-11-11
$wgLogo: Provide logo/icon/wordmark

edg2s:
As required by Vector2 and Minerva.


MatmaRex:
Vector2 has a missing image, icon.svg needs to be copied. I'll patch it.

PR #348, opened 2021-11-11
Try to use consistent casing for "Patch demo"

edg2s:
The username "Patch Demo" would probably be annoying to
change at this point.

### DIFF
--- a/includes.php
+++ b/includes.php
@@ -328,7 +328,7 @@ function delete_wiki( string $wiki ) : int {
 		$creator = $wikiData['creator'];
 		post_phab_comment(
 			'T' . $task,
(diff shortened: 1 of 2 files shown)


new/install.sh
```diff
@@ -12,7 +12,7 @@ php $PATCHDEMO/wikis/$NAME/w/maintenance/install.php \
 --scriptpath="$SERVERPATH/wikis/$NAME/w" \
 --with-extensions \
 --pass=patchdemo1 \
-"$WIKINAME" "Patch Demo"
+"$WIKINAME" "Patch demo"
```
MatmaRex, line 15: This is the username, actually.

MatmaRex (changes requested):

new/install.sh
edg2s: Fixed

MatmaRex (approved):

PR #355, opened 2021-11-12
Don't show captcha to any logged in users

### LINKED ISSUE #354: All logged in users should be able to skip ConfirmEdit
Currently on the admin account `Patch Demo` skips the captcha check, which doesn't make sense as all the accounts have the same password. We should disable the captcha for all accounts or none (I would suggest all until we detect bots logging in, which hasn't happened since we set up the captcha).

https://www.mediawiki.org/wiki/Extension:ConfirmEdit#Configuration

edg2s:
Fixes #354

### DIFF
--- a/localsettings/extensions-ConfirmEdit.txt
+++ b/localsettings/extensions-ConfirmEdit.txt
@@ -6,4 +6,7 @@ if ( file_exists( 'extensions/ConfirmEdit/QuestyCaptcha/extension.json' ) ) {
 	];
 	$wgCaptchaTriggers['edit'] = true;
 	$wgCaptchaTriggers['createaccount'] = true;
+
+	// Don't show CAPTCHA when logged in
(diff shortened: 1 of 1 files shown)


localsettings/extensions-ConfirmEdit.txt
```diff
@@ -6,4 +6,7 @@ if ( file_exists( 'extensions/ConfirmEdit/QuestyCaptcha/extension.json' ) ) {
 	];
 	$wgCaptchaTriggers['edit'] = true;
 	$wgCaptchaTriggers['createaccount'] = true;
+
+	// Don't show CAPTCHA when logged in
+	$wgGroupPermissions['user']['skipcaptcha'] = false;
```
MatmaRex, line 11: Needs to be `true` ;)

MatmaRex (changes requested):

PR #362, opened 2021-11-16
Allow users to change $wgLanguageCode

### LINKED ISSUE #361: Allow changing wiki's language
See https://phabricator.wikimedia.org/T270808#7153684

Would make testing patches like https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Scribunto/+/699521/ possible where the language has to be set to something specific.

edg2s:
Put in some simple validation to limit language codes
to letters and hyphens for security, but don't bother
to check the code is valid.

Fixes #361

### DIFF
--- a/new/postinstall.sh
+++ b/new/postinstall.sh
@@ -3,7 +3,8 @@ set -ex
 
 # update Main_Page
 sleep 1 # Ensure edit appears after creation in history
-echo "$MAINPAGE" | php $PATCHDEMO/wikis/$NAME/w/maintenance/edit.php "Main_Page"
+MAINPAGETITLE=$( echo 'echo Title::newMainPage()->getDBkey();' | php $PATCHDEMO/wikis/$NAME/w/maintenance/eval.php )
+echo "$MAINPAGE" | php $PATCHDEMO/wikis/$NAME/w/maintenance/edit.php "$MAINPAGETITLE"
 
 # run update script (#166, #244)
 php $PATCHDEMO/wikis/$NAME/w/maintenance/update.php --quick
--- a/css/common.css
+++ b/css/common.css
@@ -139,6 +139,10 @@ summary > .oo-ui-labelElement-label:not( .oo-ui-inline-help ) {
 	font-size: 0.92857143em;
 }
 
+.form-language {
+	max-width: 10em;
+}
+
 @media ( min-width: 721px ) {
 	.enableNotifications {
 		margin-left: 40%;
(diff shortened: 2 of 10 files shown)


new/postinstall.sh
```diff
@@ -1,9 +1,19 @@
...
+# Force message cache to populate
+php $PATCHDEMO/wikis/$NAME/w/maintenance/language/alltrans.php > /dev/null
+
+# Fetch main page title from message cache
+# TODO: This would be much neater if we just had a maintenance script that could access Title:newMainPage.
+MAINPAGETITLE=$( mysql \
+	-u patchdemo -ppatchdemo patchdemo_$NAME -B -N -e "SELECT lc_value FROM l10n_cache where lc_key = 'messages:mainpage' LIMIT 1" | \
+	php -r "echo unserialize(file_get_contents('php://stdin'));" \
+)
+
```
edg2s, line 13: This is horrific, but I couldn't find a better way.

We can patch maintenance/edit.php to support editing the main page, but that will never work with old versions of core, although perhaps that's ok...

edg2s: See https://gerrit.wikimedia.org/r/c/mediawiki/core/+/739555

MatmaRex: I was going to suggest this:
```
echo "{{int:mainpage}}" | php maintenance/parse.php
```
…but, that actually outputs HTML wrapped in `<p>…</p>`, so you'd need some more weird stuff to remove that.

Or, maybe we can just query api.php?action=query&format=json&meta=siteinfo and parse the "mainpage" given there?

edg2s: `{{int:mainpage}}` seems a little neater.

edg2s: Done. I've left the comment about fixing edit.php in because I think that is the best solution.

legoktm: What about:
```
echo 'echo Title::newMainPage()->getPrefixedText()' | php maintenance/eval.php 
Main Page
```
?

edg2s: Thanks, I tried eval.php but for some reason couldn't get it to read stdin.

In your example the space in "Main Page" causes it to write to the article "Main", but that should be fixable.

MatmaRex:
I tried this out on https://patchdemo.wmflabs.org/, and… the site interface for anon users stubbornly still displays in English, even though the content language is changed. I don't understand why, I guess something somewhere caches it wrong.

edg2s:
When I'm logged out the UI language is set to "British English", then "English" when I log in. Presumably we use the "Accept-Language" header for logged out users, and there is a setting for disabling this, but this is not a regression introduced by this patch.

MatmaRex:
Thanks, I completely forgot about that.

legoktm:
Thank you both!

PR #363, opened 2021-11-16
Use language code to decide which wiki to proxy from

edg2s:


edg2s:
Merge after #362

MatmaRex:
> Merge after #362

I want to see what happens if I merge this one first… WIll the other pull request be updated to include it, or what?

PR #366, opened 2021-11-18
Checkout wikis one at a time to report progress

edg2s:
Requires splitting the `checkout.sh` into three parts.

Move database creation to `install.sh`.

PR #369, opened 2021-11-19
Run `composer update` one repo at a time

edg2s:
Splits `postcheckout.sh` into `composerupdate.sh` (looped)
and `submodules.sh`, which could also be looped but for now
just fixes lib/ve.


edg2s:
Merge after #366

PR #370, opened 2021-11-19
editcounts: Add button to update site stats

edg2s:


MatmaRex:
Are the stats wrong? That might indicate some other problem.

edg2s:
Yes - the active user count does not get updated automatically (perhaps https://phabricator.wikimedia.org/T140079 ?)

PR #372, opened 2021-11-23
Make settings files complete PHP files

edg2s:


new/install.sh
```diff
@@ -14,16 +14,19 @@ php $PATCHDEMO/wikis/$NAME/w/maintenance/install.php \
 --pass=patchdemo1 \
 "$WIKINAME" "Patch Demo"
 
+mkdir $PATCHDEMO/wikis/$NAME/w/settings.d
+
+echo 'foreach( glob( __DIR__ . "/settings.d/*.php" ) as $conffile ) { include_once $conffile; }' >> $PATCHDEMO/wikis/$NAME/w/LocalSettings.php
```
edg2s, line 19: This is the technique and structure that vagrant uses for additional config files.

new/install.sh
```diff
@@ -14,16 +14,16 @@ php $PATCHDEMO/wikis/$NAME/w/maintenance/install.php \
 --pass=patchdemo1 \
 "$WIKINAME" "Patch Demo"
 
-# apply our default settings
 echo "\$wgLanguageCode = '$LANGUAGE';" >> $PATCHDEMO/wikis/$NAME/w/LocalSettings.php
-cat $PATCHDEMO/localsettings/core.txt >> $PATCHDEMO/wikis/$NAME/w/LocalSettings.php
```
edg2s, line 19: turns out we were (harmlessly) writing the core settings twice, once here, and once in the loop

PR #375, opened 2021-11-24
Set legacy $wgLogo for older releases

edg2s:
_e.g._ REL1_32


edg2s:
According to the docs in newer releases, $wgLogo is ignored when $wgLogo**s** is set.

PR #376, opened 2021-11-24
Don't use current PHP env when running shell commands

### LINKED ISSUE #374: install.php fails on REL1_30
```
+ php /var/www/html/wikis/a29030677b/w/maintenance/install.php --dbname=patchdemo_a29030677b --dbuser=patchdemo --dbpass=patchdemo --confpath=/var/www/html/wikis/a29030677b/w --server=https://patchdemo.wmflabs.org --scriptpath=/wikis/a29030677b/w --with-extensions --pass=patchdemo1 'Patch demo (REL1_30)' 'Patch Demo'
This script must be run from the command line
```

edg2s:
Fixes #374

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,8 @@
 {
 	"require": {
 		"mediawiki/oauthclient": "1.2.0",
-		"symfony/process": "^5.1",
-		"symfony/yaml": "^5.1",
+		"symfony/process": "^5.4",
+		"symfony/yaml": "^5.4",
 		"oojs/oojs-ui": "0.42.1",
 		"zolli/phabricator-php-api": "^2.0"
 	},
--- a/includes.php
+++ b/includes.php
@@ -289,7 +289,7 @@ function make_shell_command( array $env, string $cmd ): string {
 function shell_echo( string $cmd ): int {
 	echo '<pre>';
 	echo htmlspecialchars( "$cmd\n" );
-	$process = Process::fromShellCommandline( $cmd );
+	$process = Process::fromShellCommandline( $cmd, null, [] );
 	$process->setTimeout( null );
 	$error = $process->run( static function ( $type, $buffer ) {
 		echo htmlspecialchars( $buffer );
@@ -299,7 +299,7 @@ function shell_echo( string $cmd ): int {
 }
 
 function shell( $cmd ): ?string {
-	$process = Process::fromShellCommandline( $cmd );
+	$process = Process::fromShellCommandline( $cmd, null, [] );
 	$process->setTimeout( null );
 	$error = $process->run();
 	return $error ? null : $process->getOutput();


includes.php
```diff
@@ -289,7 +289,7 @@ function make_shell_command( array $env, string $cmd ): string {
 function shell_echo( string $cmd ): int {
 	echo '<pre>';
 	echo htmlspecialchars( "$cmd\n" );
-	$process = Process::fromShellCommandline( $cmd );
+	$process = Process::fromShellCommandline( $cmd, null, [] );
```
edg2s, line 292: We should probably use this $env argument instead of building the command ourselves with make_shell_command.

edg2s: See #377

MatmaRex: Make sense, the current code is from before this used `Process::fromShellCommandline()`.

MatmaRex:
This feels like a scary change, so I did some research:
* Before 2d950939bf8d4668e8d654a1e71e2f64574ba6d9, we used `system()`, which passes the current env variables to the subprocess too. The PHP docs don't state this clearly, but that's how environment variables and processes work in general, so it'd make sense for PHP to do the same.
* This article was an interesting read: https://mattallan.me/posts/how-php-environment-variables-actually-work/
* I still wasn't sure what exactly is going to happen if we prevent the subprocess from inheriting the env variables. I tried to see how MediaWiki does it, and found this: https://github.com/wikimedia/shellbox/commit/df51cf8b3b908d1b5c59be805acb030d906453fa. So it seems that this is probably the correct thing to do.

MatmaRex (changes requested):
However… this doesn't actually seem to fix #374. Does it work for you?

edg2s:
> However… this doesn't actually seem to fix #374. Does it work for you?

It works locally...

edg2s:
> > However… this doesn't actually seem to fix #374. Does it work for you?
> 
> It works locally...

I remembered I had a similar issue locally until I ran composer update to update the Process library. Doing this on the live server appears to fix the issue.

It exposes another issue which is that old releases trigger warnings in this version of PHP, which causes the installer to fail. I think we should suppress PHP warnings, as we would do on a live server.

edg2s:
The other issue is a conflict with the MAINPAGETITLE code - I think the php notices are getting pulled into that variable.

edg2s:
> The other issue is a conflict with the MAINPAGETITLE code - I think the php notices are getting pulled into that variable.

Fixed in #381

composer.json
```diff
@@ -1,8 +1,8 @@
 {
 	"require": {
 		"mediawiki/oauthclient": "1.2.0",
-		"symfony/process": "^5.1",
-		"symfony/yaml": "^5.1",
+		"symfony/process": "^5.4",
+		"symfony/yaml": "^5.4",
```
edg2s, line 5: The $env feature has been supported for ages in Process, but for some reason it wasn't working in 5.1 and is now in 5.4 (6.0 is out but requires PHP8).

MatmaRex (approved):

PR #380, opened 2021-11-25
Add ElasticSearch service

### LINKED ISSUE #356: Add ElasticSearch service
We rely on it heavily for GrowthExperiments features and would be nice to have in PatchDemo; it would make the sites more realistic for QA / beta testing, etc.

I'm not sure how hard this would be to include in the current set up. I see a couple of options:

- use a service run on WCMS
- install ElasticSearch 6 via apt-get and friends; there is prior art for scripting this in vagrant but probably not easily adaptable for this project
- Run ElasticSearch via a Docker container. This seems to be the easiest, if we're OK with installing Docker. That would also simplify the process of adding other services, too.

kostajh:
* Install Docker in VM creation
* Run ElasticSearch as a container using the image from
releng/dev-images which has WMF plugins
* Run CirrusSearch setup scripts in postinstall.sh

Periodically cleaning up old search index data is not done in this
patch. It can be done by deleting the volume associated with the
container (since this is considered throwaway data).

It would be nice if runJobs.php ran periodically for the wiki, as that's
needed to process updates to the search index. Ideas welcome on how to
do that.

Fixes #356

### DIFF
--- a/localsettings/extensions-CirrusSearch.txt
+++ b/localsettings/extensions-CirrusSearch.txt
@@ -1,0 +1,5 @@
+require_once "$IP/extensions/CirrusSearch/tests/jenkins/FullyFeaturedConfig.php";
+$wgCirrusSearchServers = [ 'localhost' ];
+$wgCirrusSearchWMFExtraFeatures = [
+	'weighted_tags' => [ 'build' => true, 'use' => true ]
+];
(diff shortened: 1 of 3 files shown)


kostajh:
warning; untested code.

MatmaRex:
Sorry I haven't responded yet, I will have a look eventually.

I'm mostly wondering how this will interact with the deduplication of the source code that patchdemo has. I don't know enough about Docker to guess what will happen.

PR #381, opened 2021-11-29
Main Page title: Ignore PHP warnings in output of eval.php

edg2s:
Warnings can be produced by maintenance scripts, e.g. when trying to build REL1_30. Ignore these when getting the main page title.

### DIFF
--- a/localsettings/core.txt
+++ b/localsettings/core.txt
@@ -1,7 +1,8 @@
 // Generic verbose error reporting stuff
+$isCli = PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg';
 error_reporting( -1 );
 ini_set( 'display_startup_errors', 1 );
-ini_set( 'display_errors', 1 );
+ini_set( 'display_errors', $isCli ? 'stderr' : 1 );
 $wgShowSQLErrors = true;
 $wgDebugDumpSql  = true;
 $wgShowDBErrorBacktrace = true;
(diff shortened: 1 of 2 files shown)


localsettings/core.txt
```diff
@@ -1,7 +1,8 @@
 // Generic verbose error reporting stuff
+$isCli = PHP_SAPI === 'cli' || PHP_SAPI !== 'phpdbg';
 error_reporting( -1 );
 ini_set( 'display_startup_errors', 1 );
-ini_set( 'display_errors', 1 );
+ini_set( 'display_errors', $isCli ? 'stderr' : 1 );
```
edg2s, line 5: the stderr setting only applies to CLI, so keep using '1' in web mode.

localsettings/core.txt
```diff
@@ -1,7 +1,8 @@
 // Generic verbose error reporting stuff
+$isCli = PHP_SAPI === 'cli' || PHP_SAPI !== 'phpdbg';
```
MatmaRex, line 2: This looks wrong, compared to code in `wfIsCLI()`.

MatmaRex (changes requested):

localsettings/core.txt
edg2s: Oops, forgot to invert the second half...

MatmaRex (approved):

PR #387, opened 2021-12-07
Restore $reposString

edg2s:
Lost during a logical conflict.


edg2s:
One branch removed all existing uses, and another branch added in some new uses...

PR #388, opened 2021-12-07
Set $wgResourceBasePath

edg2s:
This line is usually generated automatically by the installer,
but only since 1.27.

This fixes things like $wgLogo in versions <1.27.


edg2s:
e.g. https://patchdemo.wmflabs.org/wikis/38da96ac49/wiki/Main_Page

PR #389, opened 2021-12-07
Ensure core.php is always loaded first

edg2s:
Currently this works because 'core' comes before
'extension'/'feature' etc. in the alphabet, but
that is a nasty bug waiting to happen.


MatmaRex:
Wouldn't it be simpler if we had all of the configuration in just one file, eh?

PR #394, opened 2021-12-09
Prevent unwanted scroll when toggling repos

edg2s:


edg2s:
Probably an upstream issue in OOUI that a programmatic selection causes a scroll.

edg2s:
https://phabricator.wikimedia.org/T297481

MatmaRex:
I merged the upstream fix.

PR #398, opened 2021-12-11
Regularly check the user connection status when creating a wiki

### LINKED ISSUE #332: It shouldn't be possible to create a broken wiki
Whenever the setup process fails, the wiki should be deleted, however sometimes this doesn't happen:

https://patchdemo.wmflabs.org/wikis/1d4246ea64/w/
image
(Note the missing completion time)

Our error detection should be more robust.

edg2s:
This allows us to cleanly delete the wiki if the connection
is closed. Previously the wiki would just be left half-created.

We can't just use register_shutdown_function as that could fire
during a shell command and delete the wiki while the command
is still running.

Fixes #332

### DIFF
--- a/new.php
+++ b/new.php
@@ -33,6 +33,26 @@ $branchDesc = preg_replace( '/^origin\//', '', $branch );
 $creator = $user ? $user->username : '';
 $created = time();
 
+/**
+ * Check if the user has dropped their connection and delete the wiki if so
+ *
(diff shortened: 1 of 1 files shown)


MatmaRex:
Oh, GitHub baited me into resolving the merge conflict from the web interface, and now it tells me "This branch cannot be rebased due to conflicts". Very cool.

PR #402, opened 2021-12-13
Build OOUI from source if repo is included

### LINKED ISSUE #400: Support applying patches from OOUI
This would require a few steps, that should probably only run if an OOUI patch is detected:
1. Check out OOUI to some folder (mediawiki/lib/ooui)
2. Run the build script
3. Copy the files to resources/lib/ooui

While a little complicated and maybe fragile (if the build fails), this would be a big improvement in testing OOUI patches. We can only build from source if required as this may be slower too.

edg2s:
Fixes #400

### DIFF
--- a/new/postinstall.sh
+++ b/new/postinstall.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 
+PATH=$EXTRA_PATH:$PATH
+
 # run update script (#166, #244)
 php $PATCHDEMO/wikis/$NAME/w/maintenance/update.php --quick
 
@@ -26,6 +28,24 @@ for sql in $(find $PATCHDEMO/sql-perwiki -name "*.sql" -not -type d -printf '%P\
 do
 	mysql -u patchdemo -ppatchdemo patchdemo_$NAME < $PATCHDEMO/sql-perwiki/$sql
 done
+
+# OOUI build
+if [ -d $PATCHDEMO/wikis/$NAME/w/build/ooui ]; then
+	cd $PATCHDEMO/wikis/$NAME/w/build/ooui
+	npm install
+	npm x grunt build
+	cd $PATCHDEMO
+	# JS & CSS
+	cp -r $PATCHDEMO/wikis/$NAME/w/build/ooui/dist/* $PATCHDEMO/wikis/$NAME/w/resources/lib/ooui/
+	# PHP
+
+	cd $PATCHDEMO/wikis/$NAME/w
+	composer config repo.oojs/oojs-ui path build/ooui
+	# composer install has already run, so clear out the old version of OOUI
+	rm -rf vendor/oojs/oojs-ui
+	# ensure we skip the cache when re-installing
+	composer require "oojs/oojs-ui @dev" --no-cache --update-no-dev
+fi
 
 # grant FlaggedRevs editor rights to the default account
 if [ -d $PATCHDEMO/wikis/$NAME/w/extensions/FlaggedRevs ]; then
--- a/new/postinstall.sh
+++ b/new/postinstall.sh
@@ -37,14 +37,15 @@ if [ -d $PATCHDEMO/wikis/$NAME/w/build/ooui ]; then
 	cd $PATCHDEMO
 	# JS & CSS
 	cp -r $PATCHDEMO/wikis/$NAME/w/build/ooui/dist/* $PATCHDEMO/wikis/$NAME/w/resources/lib/ooui/
-	# PHP
 
+	# PHP
 	cd $PATCHDEMO/wikis/$NAME/w
 	composer config repo.oojs/oojs-ui path build/ooui
 	# composer install has already run, so clear out the old version of OOUI
 	rm -rf vendor/oojs/oojs-ui
 	# ensure we skip the cache when re-installing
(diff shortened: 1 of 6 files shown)


edg2s:
This works locally - but requires some configuring on the live site as our node version is very old.

edg2s:
Working on live now (e.g. on https://patchdemo.wmflabs.org/wikis/fbad8de3c8/w/index.php?title=Main_Page&action=edit , the new method `OO.ui.MessageWidget.prototype.onCloseButtonClick` exists), but it took a few steps:

* Install nvm (did this under user `matmarex` but could've been anywhere)
* Install node 12: `nvm install 12`
* Update npm: `npm install -g npm@latest`
* Create some directories owned by `www-data` in the home folder (`/var/www`):
  - `mkdir /var/www/.npm`
  - `mkdir /var/www/.config`

Ideally these would be incorporated into `setup.sh` but failing that documented in the readme.

new/postinstall.sh
```diff
@@ -27,6 +29,15 @@ do
 	mysql -u patchdemo -ppatchdemo patchdemo_$NAME < $PATCHDEMO/sql-perwiki/$sql
 done
 
+# OOUI build
+if [ -d $PATCHDEMO/wikis/$NAME/w/build/ooui ]; then
+	cd $PATCHDEMO/wikis/$NAME/w/build/ooui
+	npm install
+	npm x grunt build
```
edg2s, line 36: Annoyingly this also builds the PHP demos (including a composer install) but this will need to be fixed upstream.

MatmaRex:
Looks like you force-pushed f3e3525e3ebf537e6512946f1add26eacf9c406d to the wrong branch… the pull req. history says this was previously b79779c2563f735fc99e4054bce40c63f333c24d, so I'm restoring that. (And I'll merge the OOUI 0.43.0 update elsewhere.)

MatmaRex:
I think this is very cool, but the setup needs to happen in `setup.sh`, or it will be gone the next time we need to migrate to a new image (or whatever that's called).

edg2s:
> Ideally these would be incorporated into setup.sh but failing that documented in the readme.

This is the ideal, but it might not be practical to do everything completely automatically in setup.sh. I can at least document it somewhere...

new/postinstall.sh
```diff
@@ -27,6 +29,19 @@ do
...
+	cd $PATCHDEMO
+	# JS & CSS
+	cp -r $PATCHDEMO/wikis/$NAME/w/build/ooui/dist/* $PATCHDEMO/wikis/$NAME/w/resources/lib/ooui/
+	# PHP
+	cp -r $PATCHDEMO/wikis/$NAME/w/build/ooui/php/*  $PATCHDEMO/wikis/$NAME/w/vendor/oojs/oojs-ui/php/
```
Krinkle, line 41: This may be incompatible with the current non-PSR4 autoloading that OOUI-php does.

E.g. https://github.com/wikimedia/mediawiki-vendor/blob/a6e5ad22397e5f304a292fa8bb5e7ea87902c86e/composer/autoload_static.php#L2150

I don't know if it affected patchdemo, maybe it doesn't use composer with a static autoloader.

edg2s: Do you mean when testing a patch that added/removed PHP files?

Krinkle: Yes, or some other internal change that might affect how classes are loaded, inherited, traits, composed etc. Doesn't have to be a new widget per-se (which I'm guessing is less common of a change to test as it wouldn't be used).

edg2s: The majority of changes we want to test are JS/CSS only, so I'd be happy to merge this with some warning that PHP changes may not work. If you have any ideas how to fix it, I'm not familiar with the autoload stuff.

Krinkle: Ack. Maybe it's better to leave this line off then. Although if it doesn't break anything at least in the common case, I don't mind it either way.

edg2s: It looks like you can use `type:path` in your `repositories` list, and point that at the `build/ooui` folder we just created.

edg2s: Fixed. It's a little slower than a plain copy, but more robust.

jdforrester (approved):

MatmaRex:
I got it working locally with that addition, but it doesn't work on https://patchdemo.wmflabs.org. I deployed it there for now. It fails at this point:

```
+ composer require 'oojs/oojs-ui @dev' --no-cache --update-no-dev

                                           
  The "--no-cache" option does not exist.  
```

edg2s:
Yay old versions.

I found online the old way to do this was by setting COMPOSER_CACHE_DIR to /dev/null
image

..and it worked (see 3rd last line).

edg2s:
Yay old versions.

I found online the old way to do this was by setting COMPOSER_CACHE_DIR to /dev/null
image

..and it worked (see 3rd last line).

edg2s:
Yay old versions.

I found online the old way to do this was by setting COMPOSER_CACHE_DIR to /dev/null
image

..and it worked (see 3rd last line).

new/postinstall.sh
edg2s: Turns out we can use the demos, so this isn't a waste of time: #432

PR #406, opened 2021-12-14
Add `translate-manage` user right to admins

### LINKED ISSUE #404: Add translate-manage right to admins
Please add the `translate-manage` right to admins in [extensions-Translate.php](https://github.com/MatmaRex/patchdemo/blob/master/localsettings/extensions-Translate.php). I debugged on [51f093b25e](https://patchdemo.wmflabs.org/wikis/51f093b25e/wiki/Main_Page) quite a bit until I found out why I can’t see the new aggregate group form on Special:AggregateGroups… I can’t think of any negative effect of it except for making testing very orthopedic situations harder/impossible, but Patch Demo should aim at providing a usual environment, and this right is in the [default right set for translation admins on Wikimedia](https://phabricator.wikimedia.org/source/mediawiki-config/browse/master/wmf-config/CommonSettings.php;21ebb15b8a00$2955).

tacsipacsi:
This is needed for Special:AggregateGroups to allow modification.

Fixes #404

### DIFF
--- a/localsettings/extensions-Translate.php
+++ b/localsettings/extensions-Translate.php
@@ -7,3 +7,4 @@ $wgTranslateDocumentationLanguageCode = 'qqq';
 
 // Add this if you want to enable access to page translation
 $wgGroupPermissions['sysop']['pagetranslation'] = true;
+$wgGroupPermissions['sysop']['translate-manage'] = true;


tacsipacsi:
I didn’t actually test this, as setting up Patch Demo seems way too complicated compared to the change’s size, but the change seems trivial and almost impossible to break anything.

PR #413, opened 2022-01-11
Support GlobalPreferences extension

Funcer8:
GlobalPreferences is available on most Wikimedia sites, and I have a bunch of patches want to show to the public.

It has code paths for using the database of a local wiki, and it can work without CentralAuth according to my test.


MatmaRex:
Works fine, thanks!

PR #417, opened 2022-01-20
Show the current user's wikis at the top of the table

edg2s:


edg2s:
I think this view is more useful (similar to the gerrit dashboard) now that we have lots of users.

It may even remind people they have old wikis that need deleting :)

edg2s:
image

PR #419, opened 2022-01-23
Add GlobalBlocking extension

MarcoAurelioWM:
Enables the [GlobalBlocking](https://www.mediawiki.org/wiki/Extension:GlobalBlocking) extension on Patch Demo.

GlobalBlocking is in use in WMF wikis and recently it has started to receive many patches. It'd be great to test them in a test environment before they even reach e.g. the Beta Cluster or Production wikis.

PR #420, opened 2022-01-26
Grant globalblocking permissions to sysops

MarcoAurelioWM:
Follow-up to https://github.com/MatmaRex/patchdemo/pull/419
Assign to sysops all GlobalBlocking permissions for easier testing, and remove the steward user group as uneeded.


MarcoAurelioWM:
Thank you. For some reason the `unset` part isn't working. The rest is.

MatmaRex:
Oh, yeah, I don't think that works when the default value was defined in `extension.json`… Anyway, I don't care to fix it.

PR #426, opened 2022-02-07
Remove Stylelint

MatmaRex:
GitHub keeps bugging me about security issues in dependencies of
Stylelint. It has seemingly endless dependencies and seemingly endless
issues in them. It also provides minimal value, and that's being
generous. Let us excise it and salt the earth.

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,4 +1,0 @@
-composer/
-repositories/
-vendor/
-wikis/
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,3 +1,0 @@
-{
-	"extends": "stylelint-config-wikimedia"
-}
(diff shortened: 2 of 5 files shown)


edg2s:
I think it provides some value. Can you just turn off security warnings either completely of for devDependencies. Our only real dependency is OOUI (which has no other dependencies) and I think we'll know if that gets flagged with a security issue...

MatmaRex:
Yeah but _come on_ though.

MatmaRex:
This would honestly make my day.

edg2s:
I think a lot of the rules are useful, especially for third party contributions. It catches errors, duplicate code and incorrect indentation. The maintenance burden is pretty much zero in addition to eslint.

PR #434, opened 2022-02-26
Allow admins to bypass V+2 checks

edg2s:


MatmaRex:
I think this is a very elaborate access control to build on top of a very flimsy foundation… but whatever.

PR #435, opened 2022-03-01
Build Codex from source if repo is included

catrope:
Similar to OOUI, except it's JS only, no PHP.

Also add support for 'main' as an alias for the 'master' branch.
Without this, installing Codex fails, because it doesn't have a
branch named 'master'.

### DIFF
--- a/new/postinstall.sh
+++ b/new/postinstall.sh
@@ -32,7 +32,7 @@ done
 # OOUI build
 if [ -d $PATCHDEMO/wikis/$NAME/w/build/ooui ]; then
 	cd $PATCHDEMO/wikis/$NAME/w/build/ooui
-	npm install
+	npm ci
 	npm x grunt build
 	cd $PATCHDEMO
 	# JS & CSS
@@ -52,7 +52,7 @@ fi
 # TODO test
 if [ -d $PATCHDEMO/wikis/$NAME/w/build/codex ]; then
 	cd $PATCHDEMO/wikis/$NAME/w/build/codex
-	npm install
+	npm ci
 	npm run build-all
 	cd $PATCHDEMO
 	cp -r $PATCHDEMO/wikis/$NAME/w/build/codex/packages/codex/dist/* $PATCHDEMO/wikis/$NAME/w/resources/lib/codex/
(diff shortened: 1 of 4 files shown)


new/postinstall.sh
```diff
@@ -48,6 +48,17 @@ if [ -d $PATCHDEMO/wikis/$NAME/w/build/ooui ]; then
 	COMPOSER_CACHE_DIR=/dev/null composer require "oojs/oojs-ui @dev" --update-no-dev
 fi
 
+# Codex build
+# TODO test
+if [ -d $PATCHDEMO/wikis/$NAME/w/build/codex ]; then
+	cd $PATCHDEMO/wikis/$NAME/w/build/codex
+	npm install
```
edg2s, line 55: should be `npm ci`? (And in OOUI too probably)

edg2s (changes requested):

edg2s:
Do you have patches we can test with, and where they would be visible in MW?

catrope:
To test:
- Create a wiki ("A") with change 758961
- Visit wiki A with `?useskin=vector-2022`
- Test the search bar at the top of the page, which uses Codex. When you inspect the search bar, you should see CSS classes beginning with `cdx-`.
-  Create a wiki ("B") with changes 758961 and 765655. This should add Codex to the set of repos automatically, since 765655 is a change in Codex
- Visit wiki B with `?useskin=vector-2022`
- Test the search bar at the top of the page. It should work the same as on wiki A, but you should see different CSS classes. In particular, the `cdx-menu` class should be present on wiki B (on the absolutely positioned container of the typeahead suggestions) and absent on wiki A.

new/postinstall.sh
catrope: Sure. I just copied the OOUI code, but I'll change both to `npm ci`.

edg2s:
Had to use a different patch as that one is conflicting, but it appears to work: https://patchdemo.wmflabs.org/wikis/405f985b66/wiki/Main_Page?useskin=vector-2022

image

New CSS from that patch:
image

Spent a lot of time in `npm ci`, total build time was 3.5minutes

Also npm had engine warnings - I would recommended only requiring Node 12 in your repos.

MatmaRex:
Nice :o

PR #436, opened 2022-03-02
Move applypatch.sh to new

edg2s:


new.php
```diff
@@ -230,13 +230,12 @@ function set_progress( float $pc, string $label ) {
 	$patchesApplied[] = $data[0]['_number'] . ',' . $data[0]['revisions'][$revision]['_number'];
 	$commands[] = [
 		[
-			'PATCHDEMO' => __DIR__ . '/',
```
MatmaRex, line 233: The script uses this variable.

MatmaRex (changes requested):

new.php
edg2s: Bit of an unrelated change, but PATCHDEMO is part of $baseEnv so doesn't need to be repeated here

MatmaRex (approved):

PR #437, opened 2022-03-02
Repo list: Add w/ path prefix in code

edg2s:


edg2s:
This will be used by #270 but is a harmless change by itself.

Most places that use all.txt only use the keys (repo paths) and not the install path, including the one shell script that reads it directly (setup.sh).

MatmaRex:
Needs rebase

Issue #316, opened 2021-10-06
Consider enabling extensions from Wikimedia preset


edg2s:
Not all the extensions in the WMF preset are available at the moment. Here is the current list:

_Extensions_
- [x] [3D](https://www.mediawiki.org/wiki/Extension:3D) - #325
- [ ] [AdvancedSearch](https://www.mediawiki.org/wiki/Extension:AdvancedSearch)
- [ ] [AntiSpoof](https://www.mediawiki.org/wiki/Extension:AntiSpoof)
- [ ] [ApiFeatureUsage](https://www.mediawiki.org/wiki/Extension:ApiFeatureUsage)
- [ ] [Babel](https://www.mediawiki.org/wiki/Extension:Babel)
- [ ] [BounceHandler](https://www.mediawiki.org/wiki/Extension:BounceHandler)
- [ ] [Campaigns](https://www.mediawiki.org/wiki/Extension:Campaigns)
- [ ] [CentralAuth](https://www.mediawiki.org/wiki/Extension:CentralAuth)
- [ ] [CentralNotice](https://www.mediawiki.org/wiki/Extension:CentralNotice)
- [ ] [CheckUser](https://www.mediawiki.org/wiki/Extension:CheckUser)
- [ ] [CirrusSearch](https://www.mediawiki.org/wiki/Extension:CirrusSearch)
- [x] [cldr](https://www.mediawiki.org/wiki/Extension:cldr) - #470
- [ ] [CodeReview](https://www.mediawiki.org/wiki/Extension:CodeReview)
- [ ] [Collection](https://www.mediawiki.org/wiki/Extension:Collection)
- [ ] [CommonsMetadata](https://www.mediawiki.org/wiki/Extension:CommonsMetadata)
- [ ] [DismissableSiteNotice](https://www.mediawiki.org/wiki/Extension:DismissableSiteNotice)
- [ ] [Elastica](https://www.mediawiki.org/wiki/Extension:Elastica)
- [ ] [ElectronPdfService](https://www.mediawiki.org/wiki/Extension:ElectronPdfService)
- [ ] [ExtensionDistributor](https://www.mediawiki.org/wiki/Extension:ExtensionDistributor)
- [ ] [FeaturedFeeds](https://www.mediawiki.org/wiki/Extension:FeaturedFeeds)
- [ ] [FileExporter](https://www.mediawiki.org/wiki/Extension:FileExporter)
[...]

PR #451, opened 2022-03-31
Enable Flow extension

edg2s:
Part of #316


edg2s:
It would nice if this created a single empty or populated Flow board while installing, but I couldn't find an obvious way to do that. (cc @catrope ?)

PR #462, opened 2022-05-12
checkout.sh: Make sure path exists before readlink'ing it

catrope:
This avoids errors when trying to check things out into /w/build/FOO.
 
Also remove the line in precheckout.sh creating /build. This is the
wrong path, it should be /w/build, but creating that causes errors
because the "git worktree add" command wants an empty directory.

### DIFF
--- a/new/checkout.sh
+++ b/new/checkout.sh
@@ -2,6 +2,10 @@
 set -ex
 
 git --git-dir=$PATCHDEMO/repositories/$REPO_SOURCE/.git worktree prune
+
+# Make sure the path we're passing in exists, otherwise `readlink -f` will fail
+mkdir -p $PATCHDEMO/wikis/$NAME/$REPO_TARGET
+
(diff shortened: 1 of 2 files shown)


catrope:
This doesn't work either:
```
PATCHDEMO='/var/www/html' NAME='a64e1748a7' BRANCH='origin/master' REPO_SOURCE='mediawiki/core' REPO_TARGET='w/.' /var/www/html/new/checkout.sh
+ git --git-dir=/var/www/html/repositories/mediawiki/core/.git worktree prune
++ readlink -f /var/www/html/wikis/a64e1748a7/w/.
+ git --git-dir=/var/www/html/repositories/mediawiki/core/.git worktree add --detach /srv/patchdemo-wikis/a64e1748a7/w origin/master
Preparing worktree (detached HEAD 63c77060a0e)
fatal: '/srv/patchdemo-wikis/a64e1748a7/w' already exists
```

I think it complains about `/w` being non-empty, because it now contains `/w/build`.

PR #463, opened 2022-05-17
Enable Kartographer extension

edg2s:
Also enables JsonConfig as a dependency.

### DIFF
--- a/localsettings/extensions-Kartographer.php
+++ b/localsettings/extensions-Kartographer.php
@@ -1,0 +1,4 @@
+<?php
+
+$wgKartographerMapServer = 'https://kartotherian.wmflabs.org';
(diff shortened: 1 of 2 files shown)


MatmaRex:
The interface works, but the maps themselves don't display, because apparently "Domain is not allowed" (https://kartotherian.wmflabs.org/img/osm-intl,17,39.73939,-104.98491,500x500.png?lang=en&domain=patchdemo.wmflabs.org&title=Main+Page&groups=_4838087453010a8eb63e82dd291eb68814aa8cd3).

edg2s:
Huh, you'd think it would allow *.wmflabs.org.
Filed as https://phabricator.wikimedia.org/T308735

PR #465, opened 2022-05-23
enable PageTriage extension

DannyS712:
*Not* adding to wikimedia preset because its a big extension and only used on enwiki


edg2s:
enwiki also configures a draft namespace - should that be added to the config?

DannyS712:
> enwiki also configures a draft namespace - should that be added to the config?

I don't think thats necessary, at least for now

MatmaRex:
Seems to work, thanks!

PR #469, opened 2022-06-22
Add WikiLove

jdlrobson:
This has no special install requirements and is still in production.
It is going to be useful in Q1


MatmaRex:
Thanks

PR #477, opened 2022-07-13
Allow users to set the landing page when creating a wiki

### LINKED ISSUE #444: Allow users to change page linked to in the "announce" bot posting
The "announce" bit posts a link to the wiki's Main_Page, however this is often not the best place to start testing a patch.

We should add another input that takes a valid title path, e.g. "Talk:DiscussionTools" and link to that page instead.

We may also want to allow more complex URLs like "Special:UserLogin?returnto=Talk%3ADiscussionTools" for patches that require login.

edg2s:
Fixes #444

### DIFF
--- a/delete.php
+++ b/delete.php
@@ -25,7 +25,7 @@ if ( !isset( $_POST['confirm' ] ) ) {
 			( $useOAuth ? '<th>Creator</th>' : '' ) .
 		'</tr>' .
 		'<tr>' .
-			'<td data-label="Wiki" class="wiki">' . get_wiki_link( $wiki ) . '</td>' .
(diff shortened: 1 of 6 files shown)


edg2s:
Ping @MatmaRex

MatmaRex:
Man, there is way too much stuff in the form on the main page. I really feel like if you find yourself customizing this, you really need to write your own comment on Phabricator or something.

edg2s:
That's what I do currently, but it's messy to then have two comments. It's also a pain with old wikis not remembering why I created them.

The form is getting long though, maybe we can collapse an advanced section?

PR #481, opened 2022-07-31
Add DarkMode extension

### LINKED ISSUE #479: Add DarkMode extension
I know it isn't deployed at WMF, but I see a few other such extensions so hoping this is okay. Patch Demo makes it is sooo easy to test different release branches, interactions with other extensions, etc., that are otherwise a pain to do on my local. It would be a big help to have DarkMode on PatchDemo :)

There's nothing special about installation, just `wfLoadExtension( 'DarkMode' );` and that's it. It obviously should not be part of the Wikimedia preset.

Docs: https://www.mediawiki.org/wiki/Extension:DarkMode

Thanks!

edg2s:
Fixes #479


MusikAnimal:
Thank you!!! :)

PR #483, opened 2022-08-02
Phonos: improve example IPA on corpus page

MusikAnimal:
Larynx usually wants the slashes, so here we always include them.

We also only use 2-letter language codes because Larynx is limited in what it can interpret.

Some examples still don't work, especially those for unsupported languages, but this is enough for product/design to make use of Phonos on Patch Demo.

### DIFF
--- a/pages/extensions-Phonos.xml
+++ b/pages/extensions-Phonos.xml
@@ -45,87 +45,87 @@
       <text bytes="3508" sha1="60lcj0j0xism45kiuo959zs7xx2xywl" xml:space="preserve">{| class="wikitable"
 ! Word !! IPA
 |-
-| Xochimilco || {{#phonos: text=Xochimilco | ipa=sotʃiˈmilko | lang=es-MX }}
+| Xochimilco || {{#phonos: text=Xochimilco | ipa=/sotʃiˈmilko/ | lang=es }}
 |-
-| Tenochtitlan || {{#phonos: text=Tenochtitlan | ipa=tenoːt͡ʃˈtit͡ɬan | lang=es-MX }}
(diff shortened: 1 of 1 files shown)


pages/extensions-Phonos.xml
```diff
@@ -45,87 +45,87 @@
...
-| Gunwalloe || {{#phonos: text=Gunwalloe | ipa=gʌnˈwɒləʊ | lang=en-US }}
+| Gunwalloe || {{#phonos: text=Gunwalloe | ipa=/gʌnˈwɒləʊ/ | lang=en }}
 |-
-| Porthleven || {{#phonos: text=Porthleven | ipa=ˌpɔːrθˈlɛvən | lang=en-US }}
+| Porthleven || {{#phonos: text=Porthleven | ipa=/ˌpɔːrθˈlɛvən/ | lang=en }}
 |-
-| Somerset || {{#phonos: text=Somerset | ipa=ˈsʌmərsɛt | lang=en-US }}
+| Somerset || {{#phonos: text=Somerset | ipa=/ˈsʌmərsɛt/ | lang=en }}
 |}</text>
       <sha1>60lcj0j0xism45kiuo959zs7xx2xywl</sha1>
```
MatmaRex, line 130: Funny, MediaWiki doesn't care about this hash being wrong, there's not even a warning.

edg2s: My preferred technique is to edit the page, then re-export. We probably should care in case this ever enforced in the future.

MatmaRex: Deleting it would also be fine, all of the fields in the dumps seem to be optional.

PR #486, opened 2022-08-03
Create dependabot.yml

theresnotime:
- add a dependabot config which checks composer & npm dependencies
- limit to 5 PRs open at a time
- ignore patch-level updates


MatmaRex:
Eh, I don't particularly want to have to update and test stuff regularly without a good reason (like security issues, GitHub notifies me about those). I'd rather remove some dependencies than spend more time updating them (see #426).

PR #488, opened 2022-08-06
Apply patches before running installer

edg2s:
Patches could affect the outcome of the installation
process, e.g. changing extension.json dependencies.

### DIFF
--- a/new.php
+++ b/new.php
@@ -512,7 +512,22 @@ foreach ( $composerInstallRepos as $i => $repo ) {
 	$repoProgress += ( $end - $start ) / $repoCount;
 }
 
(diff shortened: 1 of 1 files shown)


edg2s:
Not sure why we did things this way around?

Could this remove the need to run update.php in postinstall?

MatmaRex:
I think I wrote it that way before you refactored how wiki creation works, when it was all in one big .sh script and it was more difficult to do it this way.

edg2s:
I wonder if we still need to run the update script after install now?

MatmaRex:
Maybe not, but it seems harmless and maybe even useful to test that path.

PR #490, opened 2022-08-10
Enable MediaSearch extension

### LINKED ISSUE #487: Please add MediaSearch
It would be great if we could add MediaSearch to this tool.

edg2s:
Not enabled by default because this is only installed on Commons.

Configure to use external API results from Commons and WikiData,
per on-wiki instructions.

Fixes #487

### DIFF
--- a/localsettings/extensions-MediaSearch.php
+++ b/localsettings/extensions-MediaSearch.php
@@ -1,0 +1,7 @@
+<?php
+
+// External search results API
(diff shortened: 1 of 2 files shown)


edg2s:
cc @josephseddon

edg2s:
minimal + MediaSearch:
https://patchdemo.wmflabs.org/wikis/4133f5da98/wiki/Special:MediaSearch?type=image&search=cats

PR #493, opened 2022-08-23
add support for Geodata

jdlrobson:
Currently getting "Repository mediawiki/extensions/GeoData not supported"
Mentioned in wikimedia but not all

(untested)


MatmaRex:
I haven't tested either, but it looks like it's spelled GeoData.

MatmaRex:
Seems to work, thanks!

PR #496, opened 2022-09-07
Update Phonos examples to use new tag syntax

edg2s:
Phonos no longer uses a parser function.


MusikAnimal:
Came here to make this PR and you had already done it for me! Thank you :)

PR #503, opened 2022-10-06
Proxy feature: Use new $wgVectorSearchApiUrl setting

catrope:
Introduced by https://gerrit.wikimedia.org/r/c/mediawiki/skins/Vector/+/838943. Normally we'd try for backwards compatibility here, but the old setting was broken anyway, so that's not needed.


MatmaRex:
Thanks!

PR #514, opened 2022-12-25
Enable the Realtime Preview feature (WikiEditor)

theresnotime:
https://mediawiki.org/wiki/Extension:WikiEditor#Realtime_Preview


MatmaRex:
Please consider changing the default value in WikiEditor, so that we won't have to maintain the configuration here. I am pretty sure this feature is ready to be enabled for anyone installing MediaWiki.

PR #517, opened 2023-01-10
Allow enabling temp user creation for IP Masking QA

kostajh:
Bug: T326651

### DIFF
--- a/new/postinstall.sh
+++ b/new/postinstall.sh
@@ -98,6 +98,11 @@ if [ "${USE_PROXY}" = "1" ]; then
 	done
 fi
 
+# Enable temporary user account creation (IP Masking)
+if [ "${USE_TEMPUSER}" = "1" ]; then
+	cp $PATCHDEMO/localsettings/feature-tempuser.php $PATCHDEMO/wikis/$NAME/w/settings.d
+fi
+
 # Enable instantCommons if selected
 if [ "${USE_INSTANT_COMMONS}" = "1" ]; then
 	cp $PATCHDEMO/localsettings/feature-instantCommons.php $PATCHDEMO/wikis/$NAME/w/settings.d
--- a/index.php
+++ b/index.php
@@ -237,6 +237,19 @@ if ( !$canCreate ) {
 						]
 					),
 					new OOUI\FieldLayout(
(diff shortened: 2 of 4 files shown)


Tchanders (commented):
This looks good to me - thanks @kostajh

new/postinstall.sh
```diff
@@ -98,6 +98,11 @@ if [ "${USE_PROXY}" = "1" ]; then
 	done
 fi
 
+# Enable temp user creation if selected
```
edg2s, line 101: "'Enable temporary user account creation (IP Masking)"

new/postinstall.sh
```diff
@@ -98,6 +98,11 @@ if [ "${USE_PROXY}" = "1" ]; then
...
+if [ "${USE_TEMPUSER}" = "1" ]; then
+  cp $PATCHDEMO/localsettings/feature-tempuser.php $PATCHDEMO/wikis/$NAME/w/settings.d
```
edg2s, line 103: use tabs

edg2s (changes requested):

MatmaRex:
Thanks for doing this!

Just for the record, I'm not a fan of adding more options to the Patch demo form – it's already so long, it barely fits on a 1080p screen. I'd usually suggest creating a "do not merge" Gerrit changeset somewhere that changes the MediaWiki defaults, and can be combined with other patches for Patch demo.

I am, however, a fan of the IP Masking project, and if this helps move it forward, then I'm for it.

image

edg2s:
I have a local patch to collapse everything below "Choose included repos" into an "Advanced" section. Thoughts @MatmaRex ?

MatmaRex:
Sounds okay to me, but I don't know how everyone who wanted these options added will feel about it ;)

kostajh:
> I'd usually suggest creating a "do not merge" Gerrit changeset somewhere that changes the MediaWiki defaults, and can be combined with other patches for Patch demo.

That seems tricky for non-technical users who want to preview or QA patches.

> I have a local patch to collapse everything below "Choose included repos" into an "Advanced" section.

That sounds like a good idea to me.

PR #518, opened 2023-01-10
Add CheckUser extension

Tchanders:


MatmaRex:
When trying to use the tools, I get an exception: "Error: Class 'Firebase\JWT\JWT' not found". I guess CheckUser might need the `composer install` step?

MatmaRex:
Alright, that makes it work. Thanks for the patch!

Tchanders:
Oh yeah, for Special:Investigate. Thanks @MatmaRex !

PR #522, opened 2023-01-16
Update PHP dependencies

edg2s:


edg2s:
It just affected adjacent lines with #520

PR #526, opened 2023-02-03
Make Vector 2022 the default skin for branches after origin/wmf/1.40.0-wmf.19

### LINKED ISSUE #525: Make Vector 2022 the default skin
Now that it is default on enwiki, the majority of users on WMF hosted sites will be seeing this skin, so we should update our config accordingly.

edg2s:
Fixes #525


edg2s:
This is ready.

PR #546, opened 2023-04-04
Add AntiSpoof extension

Tchanders:
There is a maintenance script to add existing users, but I guess they won't exist before the extension is installed


MatmaRex:
> There is a maintenance script to add existing users, but I guess they won't exist before the extension is installed

Done in #548

PR #551, opened 2023-04-18
Prefer ES6 arrow functions to anonymous functions

edg2s:


MatmaRex:
Question, does anyone actually prefer arrow functions? I think the syntax just looks ugly. Maybe it's just me.

edg2s:
I don't know, but not having to shadow `this` is an advantage

PR #555, opened 2023-05-04
Use explicit URL syntax when posting to Phabricator

edg2s:
This ensures spaces in URLs don't break them.


edg2s:
Example broken URL: https://phabricator.wikimedia.org/T54750#8813160

PR #558, opened 2023-05-07
Avoid PHP notices for private/non-existing Phab tasks

### LINKED ISSUE #553: Notice: Trying to access array offset on value of type null
Visiting https://patchdemo.wmflabs.org/ directly throws up these PHP notices

```
Notice: Trying to access array offset on value of type null in /var/www/html/includes.php on line 231

Notice: Trying to access array offset on value of type null in /var/www/html/includes.php on line 232
```

Ammarpad:
Fixes #553

### DIFF
--- a/includes.php
+++ b/includes.php
@@ -230,7 +230,7 @@ function get_task_data( int $task ): array {
 		] )->getResult();
 		
 		if ( $maniphestData === null ) {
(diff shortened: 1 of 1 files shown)


MatmaRex:
Thanks!

PR #559, opened 2023-05-08
Simplify code for handling null data from Phab API (e.g. security tasks)

### LINKED ISSUE #553: Notice: Trying to access array offset on value of type null

edg2s:


MatmaRex:
Oops, I already merged #558 and didn't notice this one. :/ They're both similar enough for me, but feel free to tweak on master if you want.

edg2s:
I think this version has less repetition.

PR #560, opened 2023-05-11
feat: add ORES extension to list of available extensions

isaranto:
- Add ORES extension in list of available extensions
- Run ORES extension maintenance script on install to populate `ores_classification` table with scores.

Bug: T319170

### DIFF
--- a/new/postinstall.sh
+++ b/new/postinstall.sh
@@ -140,5 +140,5 @@ cp $PATCHDEMO/images/favicon.ico $PATCHDEMO/wikis/$NAME/w/
 
 # Populate ORES database if selected
 if [ -d $PATCHDEMO/wikis/$NAME/w/extensions/ORES ]; then
-	php $PATCHDEMO/wikis/$NAME/w/extensions/ORES/maintenance/PopulateDatabase.php.php
+	php $PATCHDEMO/wikis/$NAME/w/extensions/ORES/maintenance/PopulateDatabase.php
 fi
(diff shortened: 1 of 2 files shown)


isaranto:
I think it would be ok as I can point it towards a dummy service (specify a different url for ores) that has a patchdemo wiki, that is only if I can know the `patchdemo_efb7ef308c` string beforehand (or ssh into the machine afterwards and change it)

MatmaRex:
It's the database name for the newly created wiki, and different for each one… Maybe that won't be a problem though. I'm not sure how complex the dummy service needs to be to work, but maybe we could just serve a file like https://patchdemo.wmflabs.org/dummy-ores.php, you'd implement whatever code is needed in there, and we'd set `$wgOresBaseUrl = "https://patchdemo.wmflabs.org/dummy-ores.php/"`, so that all ORES service queries would be served by that file? (if needed, you can get the full path in PHP code using `$_SERVER['REQUEST_URI']`, and extract the parameters from that)

isaranto:
For my purpose I can just change this in the code and instead of making a request to 
`https://ores.wikimedia.org/v3/scores/patchdemo_efb7ef308c/?models=damaging%7Cgoodfaith&revids=1%7C151&format=json`
in will make a request to 
`https://ores.wikimedia.org/v3/scores/MY_HARDCODED_WIKI_ID/?models=damaging%7Cgoodfaith&revids=1%7C151&format=json`
Since I'm not interested in the actual results just try to get some it is more than fine

MatmaRex:
I guess that's okay. This means that if you try to create a demo wiki with the ORES extension selected, but without a custom patch, it will fail. It's unselected by default though, so I don't mind that.

MatmaRex:
Closed by f5173609db13b189f415b8393e885d1d52eb6ad3 (I had some issues with squashing the pull request in the GitHub interface)

PR #561, opened 2023-05-19
Use ES6 classes for custom OOUI widgets

MatmaRex:
Just for a lark.

Do not use ES6 static properties though, as they're only introduced in ES2020, and also apparently everyone hates them.


edg2s:
We should work out the limitations of this. I did this a while ago with a personal project (https://github.com/edg2s/deal/blob/master/static/model/GameModel.js), so for example OO.mixinClass is still required.

MatmaRex:
The only limitation I noticed was that you can't set properties on `this` before calling `super`. It was trivial to avoid here, but there are some classes in OOUI and probably in Wikimedia projects where it would be more difficult. Everything else either has a direct equivalent in ES6 classes, or you can just do it by calling the OOUI methods as before.

PR #562, opened 2023-05-24
Add some configuration need for Kartographer

WMDE-Fisch:
I get the following error when trying to use patchdemo for Kartographer:

A dependency error was encountered while installing the extension "Kartographer": Could not find the registration file for the extension "JsonConfig";

I'm not very much into what's going on there, but in our dev env we use the lines I added here as a default for the Kartographer set.

### DIFF
--- a/localsettings/extensions-Kartographer.php
+++ b/localsettings/extensions-Kartographer.php
@@ -2,3 +2,19 @@
 
 $wgKartographerMapServer = 'https://kartotherian.wmflabs.org';
 $wgKartographerStaticMapframe = true;
+
+// Copied from https://www.mediawiki.org/wiki/Extension:JsonConfig#Supporting_Wikimedia_templates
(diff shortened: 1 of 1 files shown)


MatmaRex:
>A dependency error was encountered while installing the extension "Kartographer": Could not find the registration file for the extension "JsonConfig";

This probably just means that, in the "Choose included repos" section, you selected "Kartographer" but unselected "JsonConfig". However, "Kartographer" depends on "JsonConfig" (in extension.json). Patchdemo actually doesn't check these dependencies, but perhaps it should.

I can merge the change anyway, if you want?

WMDE-Fisch:
D'oh yeah seems like it. I thought I did it twice and looked out for that. But seems not. - No hard feelings about the patch then, I guess I'll close it. At least the reasoning is totally wrong. - Like I said, it was kind of a blind shot.

Thanks. Works great so far :-)

PR #537, opened 2023-03-20
Add API action 'findwikis' to find a wiki by a change number

edg2s:
Part of #285

Issue #285, opened 2021-04-10
Comment in Gerrit when a patchdemo is created


jhsoby:
The tool can leave comments in related Phabricator tasks when a patchdemo is created. It would be nice if it could leave comments on the patch itself in Gerrit as well.

edg2s:
Upstream request:
https://phabricator.wikimedia.org/T332474

PR #563, opened 2023-05-24
api.php: add CORS header

hashar:
We need the API to emit `Access-Control-Allow-Origin: *` to allow downstream usage in JavaScript application. For example the integration with https://gerrit.wikimedia.org/. Without it, the browser complains with:

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at
https://patchdemo.wmflabs.org/api.php?action=findwikis&change=922162. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

Part of #285
Follow-up #537

Bug: https://phabricator.wikimedia.org/T332474

### DIFF
--- a/api.php
+++ b/api.php
@@ -3,6 +3,7 @@
 require_once "includes.php";
 
 header( 'Content-Type: application/json' );
+header( 'Access-Control-Allow-Origin: *' );
 
 $data = null;
 


edg2s:
Deployed.

PR #571, opened 2023-06-07
Remove empty entries from get_known_pages

edg2s:


edg2s:
`extensions-Gadgets.txt` has some extra linebreaks mid-file for legibility, but these were generating blank entries in the ComboBox.

PR #575, opened 2023-06-28
Ignore dependencies on Wikimedia config repositories

### LINKED ISSUE #573: Add mediawiki/vendor repository
I am working on https://gerrit.wikimedia.org/r/c/mediawiki/extensions/UrlShortener/+/930716 which requires a new package in mediawiki/vendor that is not yet merged and must undergo security review. it'd be awesome if in the meantime we could demo it!

I'm not sure if this is as simple as allowlisting the repository? It seems like there are potential security issues just as with #19. There was talk at #112 about having a separate repository with the list of trusted users who can change sensitive things like this. Maybe if we move forward with that, the same repo could be checked for vendor changes.

MatmaRex:
Sometimes folks point 'Depends-On' entries to Wikimedia config repositories. We don't support them, but these dependencies are usually not needed to test a patch on Patch demo, so ignore them instead of erroring out.

Closes #573

### DIFF
--- a/includes.php
+++ b/includes.php
@@ -553,6 +553,10 @@ function get_repo_presets(): array {
 	return $presets;
 }
 
+function get_ignored_repos(): array {
+	return Yaml::parse( file_get_contents( __DIR__ . '/repository-lists/ignored.yaml' ) );
+}
+
 function get_known_pages(): array {
 	$pages = [
 		'Main Page'
--- a/new.php
+++ b/new.php
@@ -177,6 +177,10 @@ $linkedTasks = [];
 $commands = [];
 $usedRepos = [];
 
+$repos = get_repo_data();
+$repoCount = count( $repos );
+$ignoredRepos = get_ignored_repos();
+
 // Iterate by reference, so that we can modify the $patches array to add new entries
 foreach ( $patches as &$patch ) {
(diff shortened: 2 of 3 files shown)


MusikAnimal:
> (I deployed this on https://patchdemo.wmflabs.org/ for testing, cc @MusikAnimal)

It worked great as far as building the wiki, but it appears `composer install` wasn't ran? I'm getting `Exception caught: Class 'Endroid\QrCode\Builder\Builder' not found`

I thought a908a611c0e881c8c45f88b21116ddd617d1aa8c was supposed to make that work. Or perhaps I'm missing something else?

MatmaRex:
Ugh, it looks like we do the installation steps in the wrong order, running `composer install` before applying the patches. :/

edg2s:
Could we instead just ignore anything that isn't in all.txt? If the repo isn't there we never pull in the Depends-On. We could show a warning in the log that not all dependencies were checked out, in case it's a dependency that should be pulled it, but just hasn't been added to Patch Demo yet.

MatmaRex:
Yeah, that's a good idea, I think we could. I've been thinking in terms of "real" dependencies, where the other thing is needed to run the current thing, but that's now how we use Depends-On – it just indicates that the other patch needs to be merged before the current patch for smooth deployment in Wikimedia production, and "real" dependencies between extensions are indicated in extension.json, so failing to satisfy them will display a reasonable error anyway (but later in the process).

MatmaRex:
See #577 instead

PR #580, opened 2023-07-02
Add previously used landingPages as suggestions

edg2s:


includes.php
```diff
@@ -577,7 +579,26 @@ function get_known_pages(): array {
...
+		}
+		$stmt->bind_param( 's', $user->username );
+		$stmt->execute();
+		$res = $stmt->get_result();
+		while ( $data = $res->fetch_assoc() ) {
+			$pages[] = str_replace( '_', ' ', trim( $data[ 'landingPage' ] ) );
+		}
+	}
+	sort( $pages, SORT_NATURAL | SORT_FLAG_CASE );
+	$pages = array_unique( $pages );
```
MatmaRex, line 601: `array_unique()`, given a numeric array, returns an array with gaps, which can have surprising consequences:

```php
echo json_encode( array_unique( [ 'a', 'a', 'b', 'c' ] ) );
// => {"0":"a","2":"b","3":"c"}
```

We probably want to run `array_values()` on the result first. Or just do an `in_array()` check earlier before adding duplicates and avoid this stupid function.

edg2s: Fixed

PR #586, opened 2023-07-19
Update dependencies

edg2s:


MatmaRex:
Build failure:
```
symfony/process v6.3.0 requires php >=8.1 -> your php version (7.4.33) does not satisfy that requirement.
symfony/yaml v6.3.0 requires php >=8.1 -> your php version (7.4.33) does not satisfy that requirement.
```

edg2s:
Guess we can't do this until we update PHP

